Issue #1, opened 2024-10-30
Query does not seem to take into account embedded notes


blackgold9:
I have a note that I use for embedding all relevant content into a single place, then I export it to pdf in order to chat with. That same note when used here seems to only include the links to the content, not embedding the content itself. I think including it in the query would be a superior experience. In the example included, ![[About Me]] should be replaced with the contents of the About Me note.

[Example Note.md](https://github.com/user-attachments/files/17578528/Example.Note.md)

PR #2, opened 2024-10-31
Use MarkdownRender to render embeds in the note contents read from file.

blackgold9:
This addresses Issue #1


allenhutchison (approved):
Thanks!

PR #61, opened 2025-04-10
Add Custom Prompt Settings and Refactor Prompt Handling

A1i98:
This pull request introduces several improvements to the Gemini Scribe plugin, enhancing the flexibility and configurability of prompt management. The following changes have been implemented:

- Introduced new settings to allow users to select between default and custom prompt values, enabling a more personalized experience.
- Updated the GeminiPrompts class to utilize the new user-defined settings, ensuring custom prompts override the defaults when specified.
- Enhanced the settings UI by adding options for editing custom prompts, along with a dropdown to choose between default and custom prompt modes.
- Refactored prompt initialization across the plugin components to ensure that all prompt handling respects user settings.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -23,6 +23,15 @@ export interface ObsidianGeminiSettings {
 	chatHistory: boolean;
 	historyFolder: string;
 	showModelPicker: boolean;
+	promptMode: string;
+	customSystemPrompt: string;
+	customCompletionPrompt: string;
+	customGeneralPrompt: string;
+	customSummaryPrompt: string;
+	customRewritePrompt: string;
+	customDatePrompt: string;
+	customTimePrompt: string;
+	customContextPrompt: string;
 }
 
 const DEFAULT_SETTINGS: ObsidianGeminiSettings = {
@@ -40,6 +49,15 @@ const DEFAULT_SETTINGS: ObsidianGeminiSettings = {
 	chatHistory: false,
 	historyFolder: 'gemini-scribe',
 	showModelPicker: false,
+	promptMode: 'default',
+	customSystemPrompt: '',
+	customCompletionPrompt: '',
+	customGeneralPrompt: '',
+	customSummaryPrompt: '',
+	customRewritePrompt: '',
+	customDatePrompt: '',
+	customTimePrompt: '',
+	customContextPrompt: '',
 };
 
 export default class ObsidianGemini extends Plugin {
--- a/src/api.ts
+++ b/src/api.ts
@@ -46,7 +46,7 @@ export class GeminiApi {
 
 	constructor(plugin: ObsidianGemini) {
 		this.plugin = plugin;
-		this.prompts = new GeminiPrompts();
+		this.prompts = new GeminiPrompts(this.plugin.settings);
 		this.gemini = new GoogleGenerativeAI(this.plugin.settings.apiKey);
 	}
 
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -14,7 +14,7 @@ export class GeminiCompletions {
 
 	constructor(plugin: ObsidianGemini) {
 		this.plugin = plugin;
-		this.prompts = new GeminiPrompts();
+		this.prompts = new GeminiPrompts(this.plugin.settings);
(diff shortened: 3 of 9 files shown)


allenhutchison:
Hi Ali,

Thanks for the work on this. I'm wondering, though, if you are willing to make a few changes? If you were an early user of the plugin you might recall that there was a system very much like this in the plugin at the beginning right down to managing the prompts in the settings panel. 

I took it out for two reasons. First because it made it impossible for me to reliably fix issues that people had with the plugin. You can imagine how a custom system prompt would completely change the behaviour of the application across all commands that the plugin provides. These are difficult to debug issues because most of the time when the cropped up people fail to remember that they have modified the system prompt or the summary prompt, or the general prompt and that actually explained the behaviour they were seeing in the application.

The second reason I removed this feature was because it was difficult to use and not really expandable. It was difficult to edit a complex prompt in the setting panel, and that interface didn't provide a great way to expand on the commands the system offers.

To address this I've been working on a few things. First was a refactor of the prompt system to use templates, which allowed for variable substitution into the prompt (which is why it uses handlebars today). 

Then I refactored the history feature to store history as markdown files, which paves the way for other gemini-scribe functionality to be stored in the users vault in the same way. Next would be to give the user the ability to create new prompts in the same directory as structured markdown files, which would enable a better editing experience. 

The final piece of the plan would be to allow people to create new commands from this system by adding a dynamic command builder on load when the application scans the custom prompt directory. That would allow us to move the summary and completion prompts over to the new system as examples, and provide a path to custom chat prompts and possibly system prompts down the line.

How do you feel about going down this path? 

The first thing I would ask you to do, is to refactor the settings code to detect these prompts in a new prompts directory for gemini-scribe. The existence of a file in that directory would be the override for the existing prompt, and we could work together on how that file is structured to allow for the other functionality that I've described here. By doing this we can limit the settings code to a toggle to enable custom prompts, and we could potentially seed that directory with the default prompts from the system.

A1i98:
Hi Allen,

Thank you for your response and for creating such a great plugin. The idea you proposed for managing prompts is excellent. I deliberately avoided altering the core structure of the code as much as possible, and I needed that functionality urgently, which is why I added it.

Regarding your suggestions, I've given them some thought. If users could define folder templates similar to the Templates plugin it would allow, for instance, a custom template for the Journal folder and a different one for other folders.

If you have any further suggestions or ideas, please let me know. I would appreciate it if you could close this pull request, as I plan to work on Template Prompts and submit a new pull request.

allenhutchison:
ok, we can close this PR. I'd love to hear about your use case and what you needed to do that didn't work in the context of the existing prompts.

In thinking about this a little more last night, I think there is a really clean path for new chat prompts which is the main use case for gemini scribe.

Here is what I would propose. Let's add a custom frontmatter key, maybe something like gemini-scribe-chat-prompt and that key can point to any markdown file. That file would be used to replace the existing chat prompt. The user can keep and organize prompts wherever they want. 

In the context of the existing prompt system this new prompt could either extend or replace the generalPrompt.

A1i98:
I apologize for the delayed response; I'm managing several projects and classes at once. In my work, the report formatting varies from project to project, and I often summarize lecture notes as well as convert texts into questions. As a result, the prompts I create must produce output in very specific formats and writing styles tailored to my requirements.

One persistent challenge I've encountered is that mathematical equations frequently render incorrectly. To remedy this, I've developed a prompt that ensures equations are formatted correctly in Obsidian. Moreover, for my journals, I was in need of a swift solution to generate a plan based on today's tasks in a set format. 

I find your idea of introducing a prompt key in the frontmatter very appealing. It's truly a fantastic concept.

allenhutchison:
I'd love to get the math improvements you made. That sounds like something that should go into the default chat prompt.

I found the other day that just including a frontmatter field called: gemini_prompt: with some text was enough to get the model to start following those instructions. You might try that and see how it goes. You can also link to another note with the same field ID and the model will read in the instructions from the note.

I was a little surprised to find this just worked without any changes to the system, but in retrospect it's basically a prompt injection and I don't do anything in the system prompts to guard against this.

Issue #63, opened 2025-04-24
I'm getting literal &#x27; and &quot; sequences in my outputs


usergenic:
Seems like maybe some double encoding happening in the chain, but I get literal `&quot;` and `&#x27;` all over my outputs instead of quotes and apostrophes.

I'm running version 1.7.1 of the plugin and on Obsidian desktop in this case.

samwega:
So am I.

![Image](https://github.com/user-attachments/assets/f6cc26ce-40d7-426c-8c2a-1b9cb4a4601f)

So, it retrieves the he's "sorry" line correctly. However, when it references it again, then it gets messed up.

samwega:
Is anyone looking into this? A bit sick of having to search&replace every time. Unfortunately I don't have the chops to fix it. Thanks in advance!

allenhutchison:
@samwega so sorry to disappoint you. I work a couple of jobs and maintain this free software in my spare time.

The issue here is that I have to html render the markdown page in order for obsidian to include all embedded content in the page that I give to the model: https://github.com/allenhutchison/obsidian-gemini/blob/67774fdeb2401d3ea78ed945d2c515043d60ccc7/src/ui/gemini-view.ts#L336 

Now, even though I instruct the model to return plain text: https://github.com/allenhutchison/obsidian-gemini/blob/67774fdeb2401d3ea78ed945d2c515043d60ccc7/prompts/generalPrompt.txt#L14 it will occasionaly ignore this instruction and return html escaped text instead.

You can see here where we render the response from the model into markdown: https://github.com/allenhutchison/obsidian-gemini/blob/67774fdeb2401d3ea78ed945d2c515043d60ccc7/src/ui/gemini-view.ts#L244

I've seen this happen less with the 2.5 family of models than it did awhile back, so you might try upgrading to the newer models.

The reason for HTML rendering the context and page documents, as I said above, was originally to bring embedded content into the context of the model. That is less necessary now due to the improvements I've made in https://github.com/allenhutchison/obsidian-gemini/blob/master/src/files/file-context.ts which now includes embedded links, as well as backlinks, and links from dataview tables. So it might be time to turn that feature off and see what the results look like.

Either way, this is on my list to look at, but I'm currently working on a big refactor to move the entire system to the new version of Google's gemini sdk, which will enable some important use cases. Thanks for your paitence.

allenhutchison:
[...]

PR #81, opened 2025-05-23
Fix: Decode HTML entities in model responses

allenhutchison:
This change addresses an issue where HTML entities (e.g., &quot;, &#x27;) were displayed literally in the UI instead of their corresponding characters (", ').

A post-processing step has been added to the Gemini API response handling. The `parseModelResult` method in `GeminiApiNew` now uses a new private helper `_decodeHtmlEntities` to convert HTML entities in the markdown text content to their actual characters before the response is used by the plugin. The decoding is done using standard DOM manipulation (creating a temporary textarea element).

Unit tests have been added to `gemini-api-new.test.ts` to verify the correct decoding of various HTML entities and ensure that plain text remains unaffected.

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -1,7 +1,7 @@
 // jest.config.mjs
 export default {
   preset: 'ts-jest/presets/default-esm',
-  testEnvironment: 'node',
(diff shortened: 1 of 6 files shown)


allenhutchison:
Addresses #63

PR #231, opened 2025-11-25
Fix agent stuck in 'Running' status with hidden confirmation modal

### LINKED ISSUE #229: Agent does not write note. Status "Running" forever
I installed the new Version 4.1.0
I tried an agent with the instruction:
"Create a note with the highlights for Sirkka" 
with my "Finnland" note as base. I entered my instruction in German language ("Erstelle eine Notiz mit Highlights von Sirkka")

The result is this:
Image

Status "Running" forever. Same situation in Win11 and on Android on Obsidian 1.10.3

In the previous Version 4.0.0 a new note was created with Title "Sirkka Highlights" but without any contents...
Before I could open an issue the new version 4.1.0. appeared.

This are my Models:

Image
I have Debug Mode switched on but don't know where I find the log file.

Kind regards. Martin

allenhutchison:
Fixes #229

## Problem
Agent would hang indefinitely in "Running" status when user was asked to confirm a vault operation (like `write_file`). Investigation revealed the confirmation modal could be:
- Hidden behind the Obsidian window (especially on mobile)
- Accidentally dismissed by clicking outside it
- Lost with no way to recover it

Once the modal was gone, the agent would wait forever for a response, leaving users stuck.

## Root Cause
The `write_file` tool (and other vault operations) requires user confirmation. The code awaits `requestUserConfirmation()` which opens a modal and waits for user response. If the user never sees or loses the modal, the code waits indefinitely at this await.

### 1. Progress Indicator Updates (`execution-engine.ts:87-95`)
**Before:** Progress showed generic "Running" with no indication of what was happening  
**After:** Shows specific status:
- `"Waiting for confirmation: Write File"` - User knows they need to respond
- `"Executing: Write File"` - User knows action is in progress

This makes it immediately obvious when the agent is waiting for user input.

### 2. 60-Second Timeout (`execution-engine.ts:188-195`)
**Before:** Agent would wait forever if modal was missed  
**After:** 
- Modal auto-closes after 60 seconds
- Execution fails gracefully with clear log message
- Agent returns to ready state

Users can retry their request, and this time they'll know to look for the confirmation modal.

### 3. Improved Modal Z-Index (`tool-confirmation-modal.ts:21-23`)
**Before:** Modal could appear behind Obsidian window  
**After:** Explicit `z-index: 9999` ensures modal appears on top

This should reduce (though not eliminate) cases where modal is hidden.

### 4. Empty Response Handling (`agent-view-tools.ts:323-336`)
**Before:** If model returned empty response after tools, progress bar stayed visible forever  
**After:**
- Always hide progress bar, even with empty response
- Show helpful message: "I completed the requested actions but had trouble generating a summary..."

Prevents a related issue where progress bar doesn't hide.

### Manual Testing
- ✅ Tested write_file confirmation flow
- ✅ Verified progress indicator updates correctly
- ✅ Confirmed timeout triggers after 60 seconds
- ✅ Verified modal appears on top with z-index fix
- ✅ Tested empty response handling

### Automated Testing
```
✅ All 461 tests pass
✅ Build successful
✅ No regressions
```

## User Experience

**Before:**
```
User: "Create a note with highlights"
Agent: [Running...] (forever, no indication of what's wrong)
User: Forced to reload Obsidian
```

**After:**
```
User: "Create a note with highlights"
Agent: "Waiting for confirmation: Write File" (clear indication)
[Modal appears]
User: Clicks Allow (or waits 60s if modal is hidden)
Agent: "Executing: Write File" → completes successfully
```

## Known Limitations

This is an **interim fix**. The fundamental issue is that modals can be dismissed or hidden. The proper long-term solution is to move confirmations from modal dialogs to **in-chat UI** (similar to ChatGPT/Claude.ai), which will be addressed in a follow-up PR.

Benefits of in-chat confirmations:
- Can't be hidden or dismissed accidentally
- Always visible in conversation flow
- Better mobile experience
- More natural conversational UX

## Files Changed
- `src/tools/execution-engine.ts` - Progress updates + timeout
- `src/ui/tool-confirmation-modal.ts` - Z-index fix  
- `src/ui/agent-view/agent-view-tools.ts` - Empty response handling

## Impact
- **Fixes:** Issue #229 - agent no longer hangs indefinitely
- **Improves:** User feedback during confirmation wait
- **Adds:** Graceful timeout fallback
- **Maintains:** Full backward compatibility
- **Risk:** Low - defensive changes with clear fallbacks

### DIFF
--- a/src/tools/execution-engine.ts
+++ b/src/tools/execution-engine.ts
@@ -184,9 +184,13 @@ export class ToolExecutionEngine {
 	): Promise<{ confirmed: boolean; allowWithoutConfirmation?: boolean }> {
 		return new Promise((resolve) => {
 			let modalInstance: ToolConfirmationModal | null = null;
+			let resolved = false; // Prevent double-resolution race condition
 
 			// Add 60 second timeout
 			const timeoutId = setTimeout(() => {
+				if (resolved) return; // Already resolved by user
+				resolved = true;
+
 				if (modalInstance) {
 					modalInstance.close();
 				}
@@ -199,6 +203,9 @@ export class ToolExecutionEngine {
 				tool,
 				parameters,
 				(confirmed, allowWithoutConfirmation) => {
+					if (resolved) return; // Already resolved by timeout
+					resolved = true;
+
 					clearTimeout(timeoutId); // Clear timeout on user response
 					resolve({ confirmed, allowWithoutConfirmation: allowWithoutConfirmation || false });
 				}


allenhutchison:
Thank you for the thorough review! I've addressed the race condition issue and wanted to provide an update on each observation:

## Addressed in Latest Commit (6ec6c58)

**✅ #2 - Modal Close Timing Race Condition**

Added a `resolved` flag to prevent double-resolution if the user clicks during the timeout window:

```typescript
let resolved = false; // Prevent double-resolution race condition

const timeoutId = setTimeout(() => {
    if (resolved) return; // Already resolved by user
    resolved = true;
    // ... timeout logic
}, 60000);

modalInstance = new ToolConfirmationModal(
    this.plugin.app,
    tool,
    parameters,
    (confirmed, allowWithoutConfirmation) => {
        if (resolved) return; // Already resolved by timeout
        resolved = true;
        clearTimeout(timeoutId);
        // ... callback logic
    }
);
```

This ensures only one path (timeout OR user action) can resolve the Promise.

## Remaining Observations

**#1 - Timeout Configuration**: Good suggestion. Since this is an interim solution and will be replaced by in-chat confirmations (PR #232), I'd prefer to keep it simple for now rather than add settings that will be removed soon.

**#3 - Empty Response Message**: The current message does mention "completed the requested actions" to indicate execution happened. Could improve to list specific tools, but again, this will be replaced by the in-chat UI which provides much better visibility.

**#4 - Z-Index Value**: Valid point. I checked and Obsidian modals use z-index values in the range of 100-500. However, since this modal is temporary (replaced by in-chat confirmations), and 9999 ensures visibility for the interim period, I think it's acceptable for now.

**#5 - Test Coverage**: Agreed that timeout behavior should be tested. Given this is temporary code, I opted to rely on manual testing. The replacement in-chat confirmation system (PR #232) has comprehensive test coverage.

## Answers to Questions

1. **Mobile Testing**: Tested on desktop. The in-chat confirmation system (PR #232) provides the real mobile fix since confirmations can't be hidden or dismissed.

2. **In-Chat Timeline**: Already implemented! PR #232 is ready for review and provides the complete replacement for modal-based confirmations.

3. **User Configuration**: See answer to #1 - keeping simple since it's temporary.

## Summary

The critical race condition is now fixed. The remaining observations are valid but low-priority given this code will be replaced by PR #232. I believe it's ready to merge as an interim fix while the proper solution undergoes review.

Thanks again for the detailed feedback!

PR #305, opened 2026-01-22
fix(ui): prevent message submission during IME composition

shijincheng0223-dot:
## Summary
Fixes an issue where pressing `Enter` during IME composition (e.g., when typing Chinese, Japanese using Pinyin/Kana) would prematurely submit the message to the agent.
## Changes
- Added a check for `e.isComposing` in the input field's `keydown` handler.
- Prevents `callbacks.sendMessage()` triggers while the user is still composing text via IME.
## Test Plan
- Type in the chat input using an IME (e.g., MacOS Pinyin).
- Press `Enter` to select a candidate word.
- Verify that the message is NOT sent.
- Press `Enter` again after composition is finished.
- Verify that the message IS sent.

## Bug Fixes

* Fixed unintended input submission when using IME (Input Method Editor) input methods during text composition in input fields.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/ui/agent-view/agent-view-ui.ts
+++ b/src/ui/agent-view/agent-view-ui.ts
@@ -54,7 +54,7 @@ export class AgentViewUI {
 	constructor(
 		private app: App,
 		private plugin: ObsidianGemini
-	) {}
+	) { }
 
 	/**
 	 * Creates the main agent interface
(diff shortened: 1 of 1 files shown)


allenhutchison (approved):
Thank you for the contribution.

PR #595, opened 2026-04-10
feat: Add curated models.json and GitHub Action for model updates

allenhutchison:
## Summary

Stage 1 of replacing dynamic model discovery with a curated model list. Adds the JSON model list, update script, and GitHub Action — no runtime code changes.

## Changes

- Add `src/data/models.json` seeded with `*-latest` aliases as defaults
- Add `scripts/update-models.mjs` to detect new Gemini models from Google's API
- Add `.github/workflows/model-update.yml` to run weekly and open PRs when new models appear
- Filters out deprecated (gemini-2.0), specialized (robotics), and non-generative models

### Required

- [x] I have read and agree to the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] I have read and agree to the [AI Policy](../AI_POLICY.md)
- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`)
- [x] I have tested this change on Desktop
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — N/A, no runtime changes
- [x] Documentation has been updated (if applicable) — N/A, infrastructure only
- [x] I understand that I must address all review comments from CodeRabbit and maintainers, or this PR may be closed

### AI-Generated Code

- [x] This PR includes AI-generated or AI-assisted code
- [x] AI tool(s) used: Claude Code
- [x] I have reviewed and understand all AI-generated code in this PR

## Summary by CodeRabbit

* **New Features**
  * Added new Gemini model entries, including image-capable model (gemini-2.5-flash-image) and updated defaults for gemini-pro/flash variants.

* **Chores**
  * Enabled automated weekly checks (and manual runs) that detect new Gemini models and open update proposals for review and curation.

PR #596, opened 2026-04-10
refactor: Replace model discovery with curated JSON model list

allenhutchison:
## Summary

Stages 2+3: Remove the runtime model discovery pipeline and replace it with a `ModelListProvider` backed by the curated `models.json` from #595.

**Net result: -2,085 lines** of fragile discovery/filtering code replaced with a simple JSON-backed provider.

## Changes

- Create `src/services/model-list-provider.ts` — bundles models.json at build time, fetches updates from GitHub raw URL with 24h cache
- Simplify `model-manager.ts` from ~470 lines to ~140 lines
- Update `parameter-validation.ts` to use `GeminiModel` instead of `GoogleModel`
- Remove `ModelDiscoverySettings` from settings interface and defaults
- Remove model discovery UI section (toggle, interval, refresh button)
- Add settings migration to strip deprecated `modelDiscovery` data
- Delete `model-discovery.ts`, `model-mapper.ts`, `verify-model-metadata.js`
- Update all affected tests

**Depends on #595** — base branch is `feat/curated-model-list`.

### Required

- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards)
- [x] Documentation has been updated (if applicable) — settings docs need updating (model discovery removal)

### AI-Generated Code

## Summary by CodeRabbit

* **Refactor**
  * Removed dynamic model discovery and related runtime/mapping logic; model handling now relies on a bundled baseline.

* **New Features**
  * Added optional remote model synchronization to refresh the bundled model list periodically.

* **UI**
  * Removed the Model Discovery section from API settings.

* **Bug Fixes**
  * Simplified model management and initialization to improve reliability and reduce discovery-related failures.


allenhutchison:
Addressed CodeRabbit review feedback in 77ab190:

1. **\`settings-helpers.ts\` — guard the manager instance** — Now calls \`plugin.getModelManager?.()\` and guards the returned value. The fallback branch now does role-aware filtering on \`GEMINI_MODELS\` (image vs text) instead of returning the unfiltered list.

2. **\`lifecycle-service.ts\` — await syncModels()** — \`syncModels()\` is now awaited in \`initializeCoreServices()\` so subsequent startup steps see reconciled model settings.

3. **\`main.ts\` — don't run updateModelVersions() too early** — Removed the call from \`loadSettings()\` (and deleted the now-unused method). Reconciliation now happens exclusively in \`LifecycleService.syncModels()\`, which runs after \`ModelListProvider\` has loaded the cached remote model list. Previously the early call would check \`imageModelName\` against \`DEFAULT_GEMINI_MODELS\` before the remote cache was hydrated, which could silently migrate the setting.

4. **\`model-list-provider.ts\` — don't clobber settings** — Added \`remoteModelCache?: { models; timestamp }\` to \`ObsidianGeminiSettings\`. The provider now reads from \`plugin.settings.remoteModelCache\` in \`initialize()\` and writes via \`plugin.settings\` + \`saveData(plugin.settings)\` in \`fetchRemoteModels()\`. This goes through the same in-memory settings object as \`ObsidianGemini.saveSettings()\`, avoiding any clobber risk.

**Nitpick — barrel-level break in \`src/index.ts\`:** Acknowledged but not addressing here. The removed exports (\`ModelDiscoverySettings\`, \`GoogleModel\`, \`ModelDiscoveryResult\`, \`ModelDiscoveryService\`) were types/classes for the dynamic discovery system being removed entirely; there's no new module to point consumers at. This should be called out in the release notes when this PR ships, which I'll leave to the maintainer.

allenhutchison:
Addressed CodeRabbit feedback in a53d13a:

1. **Reject non-finite parameter inputs** — \`validateTemperature\` and \`validateTopP\` now early-return with a clear warning when given \`NaN\` or \`Infinity\`. Added test coverage for both cases.

2. **Simplify redundant nullish check** — Changed \`models && models.length > 0\` to \`models.length > 0\` since \`models\` is a required array parameter.

**Regarding the earlier "global max temperature" comment:** The temperature slider in settings is a single global value that applies to all model calls (chat, summary, completions). Using the global max across the catalog for the slider bounds is intentional — it's the most permissive range so any selected model can reach its cap. Per-model enforcement already happens in \`ParameterValidationService.validateTemperature()\` when a \`modelName\` is passed (e.g., for per-request validation at call time). The settings-level slider doesn't know which model will be used for which call, so it can't pick a single model's cap without being incorrect for the others.

In practice all current Gemini text models cap at 2, and image models have a lower cap but don't use temperature the same way, so this distinction is moot today — but the split is intentional design, not a bug.

Issue #634, opened 2026-04-14
Agent-initiated background execution for deep-research and image-gen


allenhutchison:
Parent: #589
Depends on: #632

Implements PR 1b of the background/scheduled tasks work.

## Scope

- [ ] Add optional `background: boolean` parameter to `DeepResearchTool` (`src/tools/deep-research-tool.ts`) and `GenerateImageTool` (`src/tools/image-tools.ts`); default `false` for backward compatibility
- [ ] When `background: true`, submit to `BackgroundTaskManager` and return `{ taskId, outputPath }` synchronously instead of awaiting the work
- [ ] Agent retrieves results later by reading the file at the output path it specified — no new tool call needed, existing `read_file` suffices
- [ ] Update `prompts/agentToolsPrompt.hbs` so the agent understands when to use `background: true` (long-running research/image operations mid-plan)

## Non-goals

- Session-resume completion banners — decided against (status bar Notice + output file are sufficient)
- Auto-notifying the agent when a background task finishes — fire-and-forget only

## Tests

- Tool execution with `background: true` returns immediately with task ID
- Output path correctly passed through to the background task
- Backward compatibility: omitting `background` behaves as today

Issue #633, opened 2026-04-14
Scheduled task execution engine: manager, markdown-per-task storage, headless session runner


allenhutchison:
Parent: #589
Depends on: #632 (shares background infrastructure)

Implements PR 2a of the background/scheduled tasks work.

## Scope

- [ ] `ScheduledTaskManager` service with a 60s `setInterval` loop (same pattern as `ChatTimer`)
- [ ] **Task definitions** stored as markdown files in `{historyFolder}/Scheduled-Tasks/<slug>.md`:
  - Frontmatter: `schedule`, `enabledTools[]`, `outputPath` (template, may include `{date}` etc.), `enabled`, `runIfMissed`
  - Body: the prompt text
  - Matches existing patterns (custom prompts, skills, agent sessions all live as markdown)
- [ ] **Runtime state** in sidecar `{historyFolder}/Scheduled-Tasks/scheduled-tasks-state.json`: `nextRunAt`, `lastRunAt`, `lastError` per task slug. Kept separate so scheduler ticks don't churn user-authored markdown frontmatter. Descriptive non-hidden name matches existing patterns (`example-prompts.json`, `rag-index-cache.json`).
- [ ] Pure `computeNextRunAt(schedule, from)` — supports `once`, `daily`, `weekly`, `interval`; unit-tested in isolation
- [ ] Scheduled task execution = headless agent session with the task's prompt + tools; output lands at the configured `outputPath` (typically `Scheduled-Tasks/Runs/<slug>/<date>.md`)
- [ ] Task discovery on layout-ready via `vault.getMarkdownFiles()` filtered by path prefix

## Layout

```
gemini-scribe/Scheduled-Tasks/
├── <task-slug>.md                    ← definition (user-edited)
├── Runs/
│   └── <task-slug>/
│       └── <date>.md                 ← result output
└── scheduled-tasks-state.json        ← volatile runtime state
```

## Non-goals

- Catch-up UX for missed runs — follow-up issue
- CRUD modal / commands — follow-up issue

## Tests

- `computeNextRunAt` for each schedule type
- Task discovery from markdown files
- Sidecar state read/write
- Headless session execution produces expected output file

PR #637, opened 2026-04-14
Feat/background task infrastructure

### LINKED ISSUE #589: Scheduled and background tasks
It would be neat to have scheduled tasks. 

Also some tasks like deep research and image generation should be run as background tasks.

saheersk:
## Summary

Implements PR 1a of the background/scheduled tasks work discussed in #589.

Adds a `BackgroundTaskManager` service so long-running operations (deep research, image generation) can run without blocking the UI. Includes a single coordinated status bar indicator, a monitoring modal, and completion notices with vault links. Also adds an `output_path` parameter to `GenerateImageTool` so background image submissions have a predictable, addressable output location.

Fixes #589

## Changes

- **`BackgroundTaskManager` service** — `submit(type, label, work)` returns a task ID immediately; caller never blocks. Emits `backgroundTaskStarted`, `backgroundTaskComplete`, and `backgroundTaskFailed` events on the existing `AgentEventBus`
- **`BackgroundStatusBar`** — single coordinated status bar item reflecting both RAG indexing state and background task state (not two icons side by side). Click opens the monitoring modal
- **`BackgroundTasksModal`** — lists running tasks with Cancel button and the 20 most recent completed/failed/cancelled tasks, each with an Open result link when an output file exists
- **`output_path` on `GenerateImageTool`** — optional explicit vault path for the output file so fire-and-forget image submissions are locatable after the fact
- **`styles.css`** — modal task row layout (icon, label, meta on separate lines, status colours), Cancel button styling, status bar spin animation
- **`AgentEventBus` event types** — added `backgroundTaskStarted`, `backgroundTaskComplete`, `backgroundTaskFailed` to `agent-events.ts`
- **`LifecycleService`** — registers `BackgroundTaskManager` and `BackgroundStatusBar` once, alongside the event bus, surviving re-initialisation
- **Command** — "View Background Tasks" added to the command palette
- **Tests** — task lifecycle transitions (pending → running → complete/failed/cancelled), cancellation, event emission, `output_path` forwarding
- **Docs** — `docs/guide/background-tasks.md` describing the status bar, monitoring modal, and result access

## Non-goals (deferred to follow-up issues)

- Agent `background: true` flag — see #634
- Dedupe of identical in-flight submissions
- Shared `BackgroundWork` interface with RAG
- Scheduled task engine — see #633 (next in sequence)

## Screenshots / Screencast

**Background Tasks modal — running task with Cancel:**
Modal shows spinning loader icon, task label, start time, and a red Cancel button. Cancellation works immediately and moves the task to the Recent section.

**Background Tasks modal — completed tasks:**
Completed tasks show a green check icon. Tasks with output files show an Open result link that navigates to the note in Obsidian.

**Status bar indicator:**
Single indicator at the bottom of the screen. Shows spinning icon + task count while tasks are running. Disappears when idle.

### Required

### AI-Generated Code

## Next Steps

After this merges, #633 will implement the scheduled task execution engine — `ScheduledTaskManager` with a 60s setInterval loop, markdown-per-task storage in `{historyFolder}/Scheduled-Tasks/`, headless agent session execution, and pure `computeNextRunAt()` for unit-tested scheduling logic.

## Summary by CodeRabbit

* **New Features**
  * Unified Background Tasks status indicator combining background work and RAG indexing progress
  * Background Tasks panel: view running tasks (with per-task Cancel) and recent history; "Open result" links for completed outputs
  * Optional output_path to control image generation save location

* **Documentation**
  * New guide describing background task behavior, how to open the panel, cancellation semantics, and locating results

* **Behavior**
  * Completion/failure notices include clickable vault links and clear statuses

### DIFF
--- a/docs/guide/background-tasks.md
+++ b/docs/guide/background-tasks.md
@@ -11,7 +11,7 @@ The status bar shows a single indicator for all background work. It reflects bot
 | Hidden                         | No background work running, RAG idle or disabled |
 | Spinning loader icon + count   | One or more background tasks running             |
 | Upload-cloud icon + percentage | RAG indexing in progress                         |
-| Database icon + number         | RAG idle, N files indexed                        |
+| Pause-circle icon              | RAG indexing paused                              |
 
 Click the indicator at any time to open the **Background Tasks** panel.
 
--- a/docs/guide/background-tasks.md
+++ b/docs/guide/background-tasks.md
@@ -1,0 +1,57 @@
+# Background Tasks
+
+Long-running operations — deep research and image generation — run in the background so they never block your editing session. Results are saved to your vault automatically and you're notified when they're ready.
+
+## Status Bar Indicator
+
+The status bar shows a single indicator for all background work. It reflects both active background tasks and the RAG indexing state in one place.
+
+| Appearance                     | Meaning                                          |
+| ------------------------------ | ------------------------------------------------ |
+| Hidden                         | No background work running, RAG idle or disabled |
+| Spinning loader icon + count   | One or more background tasks running             |
+| Upload-cloud icon + percentage | RAG indexing in progress                         |
+| Database icon + number         | RAG idle, N files indexed                        |
+
+Click the indicator at any time to open the **Background Tasks** panel.
+
+## Background Tasks Panel
+
+The panel (also available via **Command Palette → View Background Tasks**) shows:
+
+- **Running** — tasks currently in progress, with a Cancel button for each
+- **Recent** — the last 20 completed, failed, or cancelled tasks
+
+Completed tasks with output files show an **Open result** link that opens the file in Obsidian.
+
+## How Tasks Are Created
+
+Background tasks are created automatically when you trigger long-running operations:
+
+- **Deep Research** — starts a research task; result saved to `<history-folder>/background-tasks/<id>.md`
+- **Image Generation** — generates and saves an image; result path shown in the completion notice
+
+Both operations fire a completion notice with a clickable vault link when done. If a task fails, a notice explains the error.
+
+## Cancellation
+
+Click **Cancel** next to a running task in the panel. The task stops at the next safe checkpoint — it may not stop instantly if the underlying API call is already in flight.
+
+## Accessing Results
+
+When a task completes you'll see a notice in the bottom-right corner with an **Open result** link. You can also find the result by:
+
+1. Clicking the status bar indicator → **Open result** in the Recent section
+2. Navigating directly to `<history-folder>/background-tasks/` in the file explorer
+
+## Troubleshooting
+
+**Task shows as failed**
+Check the error shown in the Background Tasks panel. Common causes:
+
+- API key not configured or expired
+- Network timeout on long-running research queries
+- Vault path conflict for image output
+
+**Status bar indicator not visible**
+The indicator is hidden when there is nothing to show (no tasks running, RAG disabled or idle). Trigger a background task or enable RAG indexing in Settings.
--- a/src/main.ts
+++ b/src/main.ts
@@ -36,6 +36,8 @@ import { ProjectManager } from './services/project-manager';
 import { AgentEventBus } from './agent/agent-event-bus';
 import { ToolExecutionLogger } from './subscribers/tool-execution-logger';
 import { LifecycleService } from './services/lifecycle-service';
+import { BackgroundTaskManager } from './services/background-task-manager';
+import { BackgroundStatusBar } from './services/background-status-bar';
 import { getErrorMessage } from './utils/error-utils';
 
 export interface RagIndexingSettings {
@@ -182,6 +184,8 @@ export default class ObsidianGemini extends Plugin {
 	public projectManager!: ProjectManager;
 	public agentEventBus!: AgentEventBus;
 	public toolExecutionLogger: ToolExecutionLogger | null = null;
+	public backgroundTaskManager: BackgroundTaskManager | null = null;
+	public backgroundStatusBar: BackgroundStatusBar | null = null;
 
 	// Private members
 	private ribbonIcon!: HTMLElement;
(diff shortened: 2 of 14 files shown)


allenhutchison (commented):
Thanks @saheersk — this is a really strong first contribution. The scope matches what we agreed in #589, the lifecycle integration is correct, and the test coverage hits the right surface. Tests pass cleanly locally (43/43). A few things to address before we merge, then a couple I'd just file as follow-ups.

## Please fix in this PR

**1. `destroy()` doesn't actually stop in-flight work** — `src/services/background-task-manager.ts:447-454`

```typescript
destroy(): void {
    for (const task of this.getActiveTasks()) {
        this.cancellationFlags.set(task.id, true);
    }
    this.tasks.clear();
    this.cancellationFlags.clear();   // ← immediately wipes the flag we just set
}
```

`cancellationFlags.clear()` runs before any in-flight `run()` polls `isCancelled()`, so the work proceeds, mutates a dangling task reference, **emits `backgroundTaskComplete`**, and **shows a completion Notice** after the plugin has been told to tear down. Settings changes trigger a full plugin reload here, so this is reachable in normal use — users would see ghost notices. Drop the `cancellationFlags.clear()` line and let the existing `finally { this.cancellationFlags.delete(id) }` handle cleanup.

**2. `output_path` needs path validation** — `src/services/image-generation.ts:144-150`

The new `output_path` parameter is forwarded straight to `vault.createBinary()` with no normalization or sandboxing. Once #634 lands and the agent can call this autonomously, a prompt-injected agent could write to `../foo.png`, `.obsidian/themes/x.png`, or inside the plugin's state folder. Per the System Folder Protection guidance in AGENTS.md, the explicit-path branch should:
- pass through `normalizePath()`
- reject paths under `settings.historyFolder` or `.obsidian/`
- reject paths that escape the vault

Better to bake this in here than retrofit it under #634 when the attack surface opens up.

**3. Status bar has an unreachable branch and a doc/code mismatch** — `src/services/background-status-bar.ts:170-216` vs. `docs/guide/background-tasks.md:15-20`

The early return hides the bar whenever `runningCount === 0 && (ragStatus === 'idle' || 'disabled')`, which makes the `} else if (ragStatus === 'idle' && runningCount === 0)` branch later in `update()` unreachable. The docs row `\"Database icon + number — RAG idle, N files indexed\"` describes a state that never renders. Pick one — either remove the dead branch and that docs row, or render an idle-but-visible state. I'd lean toward removing them; the bar staying hidden when there's nothing happening is the right behavior.

## Optional / fine to defer to follow-up

- **Throw-during-cancel inconsistency** (`background-task-manager.ts:400-416`): when work is cancelled and then throws, the emitted `backgroundTaskFailed` event carries the throw's error message rather than `'Cancelled'`, while the clean-cancel path emits `error: 'Cancelled'`. Subscribers see two shapes for one logical outcome. Worth a one-liner if you're touching the file anyway, but not blocking.
- **`runningCount` test is non-deterministic** (`background-task-manager.test.ts:1240-1255`): `expect(manager.runningCount).toBeGreaterThanOrEqual(0)` is trivially true. Either remove or await one micro-task tick after submit so `runningCount === 1` is reliable.
- **Dead `RagStatusBar`**: now that `setup()` is never called on it, the class and its `update()` call from `RagIndexingService.updateStatusBar()` are effectively dead. Happy to take this as a separate cleanup PR — out of scope here.

## OK as-is

- Modal re-render via `this.onOpen()` after Cancel — wasteful but fine for low-frequency ops.
- `style.display` toggling vs. a CSS class — style preference, not worth churning.
- In-memory `nextId` reset on reload — fine for ephemeral tasks; a one-line comment noting that would be nice but not required.
- Completion Notice DOM construction — safely uses `createSpan`/`createEl`, no innerHTML, good.

## One more thing

The mobile checkbox is unchecked, which is honest. Could you do a quick smoke test on mobile for status-bar visibility and modal layout before we merge? Nothing in the implementation looks platform-specific, so I expect it's fine, but worth confirming.

Nice work overall — this lines up cleanly for #634/#633.

saheersk:
Hey @allenhutchison, thanks for the thorough review! All required fixes and the optional ones are addressed in the latest push. Here's a summary of what was done:

**Required fixes**

1. **`destroy()` ghost notices** — removed `cancellationFlags.clear()` so in-flight work sees `isCancelled() === true` after teardown. The existing `finally { cancellationFlags.delete(id) }` handles per-task cleanup as intended.

2. **`output_path` validation** — added `validateOutputPath()` to `ImageGeneration` that runs `normalizePath()` and rejects vault-escaping paths (`..`), `.obsidian/`, and the plugin state folder. Follows the same pattern as `validateAndNormalizeFilePath()` in `deep-research.ts`.

3. **Unreachable branch + docs mismatch** — removed the dead `else if (ragStatus === 'idle' && runningCount === 0)` branch from `BackgroundStatusBar.update()` and dropped the matching "Database icon + number" row from `docs/guide/background-tasks.md`. Bar stays hidden when nothing is happening.

**Optional fixes (done while in the file)**

4. **Throw-during-cancel** —`catch` block now always emits `error: 'Cancelled'` when `isCancelled()` is true, consistent with the clean-cancel path.

5. **Non-deterministic `runningCount` test** — replaced `toBeGreaterThanOrEqual(0)` with `toBe(1)`. Reliable because `run()` sets `status = 'running'` synchronously before its first `await`.

**Mobile smoke test**

Tested via `emulateMobile(true)` on desktop. Background Tasks modal opens and renders correctly on mobile layout. Console errors visible are pre-existing NodeJS package warnings from MCP code, unrelated to this PR — happy to file a separate issue for that if useful.

**One thing to flag** — now that `BackgroundStatusBar` absorbs RAG state, `RagStatusBar` and the `updateStatusBar()` call in `RagIndexingService` are dead code (`setup()` is never called, so `statusBarItem` is always null). Left it in place as you suggested will send a follow-up cleanup PR to remove `rag-status-bar.ts` and the dead call once this merges.

Issue #632, opened 2026-04-14
Background task infrastructure: manager, status bar, monitoring modal, image-gen output path


allenhutchison:
Parent: #589

Implements PR 1a of the background/scheduled tasks work.

## Scope

- [ ] New `BackgroundTaskManager` service registered via `LifecycleService` (alongside existing services)
- [ ] `submit(type, label, work)` returns a task ID immediately; caller never blocks
- [ ] Emits `backgroundTaskStarted` / `backgroundTaskComplete` / `backgroundTaskFailed` events on the existing `AgentEventBus`
- [ ] Status bar icon — single, coordinated surface that reflects both RAG state and background task state (not two icons side by side). Click opens the monitoring modal.
- [ ] "View Background Tasks" modal: lists running + recent tasks, allows cancellation
- [ ] `Notice` with vault link to the output file on completion
- [ ] Add optional `output_path` parameter to `GenerateImageTool` (`src/tools/image-tools.ts`) so background submissions are addressable — without this, a fire-and-forget image gen has no way to be located after the fact. Deep research already has this via `outputFile` (`src/tools/deep-research-tool.ts:38`).

## Non-goals

- Agent `background: true` flag — see follow-up issue
- Dedupe of identical in-flight submissions — deferred
- Shared `BackgroundWork` interface with RAG — deferred until RAG actually migrates

## Tests

- Task lifecycle transitions (pending → running → complete/failed/cancelled)
- Cancellation
- Event emission

## Docs

- Section in `docs/guide/` describing the status bar surface and the monitoring modal

allenhutchison:
Shipped in #637 (Apr 16) — `BackgroundTaskManager`, status bar, monitoring modal, and image-gen `output_path` parameter all landed there. Closing as part of the epic-completeness audit; the sub-issue was never closed because #637 referenced only the parent #589.

PR #651, opened 2026-04-18
Feat/background tool mode

### LINKED ISSUE #634: Agent-initiated background execution for deep-research and image-gen

saheersk:
## Summary

Implements PR 1b of the background/scheduled tasks work (parent: #589, depends on: #632).

Adds an optional `background: boolean` parameter to `DeepResearchTool` and `GenerateImageTool`. When `background: true`, the tool submits work to `BackgroundTaskManager` and returns `{ taskId, output_file }` or `{ taskId, output_path }` immediately — without blocking the agent turn. The agent retrieves results later by calling `read_file` on the returned path. Default is `false` for full backward compatibility.

Closes #634

## Changes

- `src/tools/deep-research-tool.ts`: add `background` parameter; background path submits to `BackgroundTaskManager`; auto-generates output path as `Background Research/YYYY-MM-DD <topic>.md` at vault root using `formatLocalDate()` + `sanitizeFileName()`; cancel poller wires up `DeepResearchService.cancelResearch()` every 2 s when `isCancelled()` flips true
- `src/tools/image-tools.ts`: add `background` parameter; captures active file path at submit time so attachment-folder resolution is correct even if active file changes while task is in flight; returns `output_path: null` when no explicit path is given — image lands in vault's configured attachment folder via `getAvailablePathForAttachment`; agent uses `find_files_by_name` to locate it
- `src/services/background-status-bar.ts`: fix regression — status bar now hides only when RAG is `disabled` AND no tasks running; RAG idle state (database icon + indexed-file count) is correctly shown when no tasks are running
- `src/ui/background-tasks-modal.ts`: subscribe to `backgroundTaskStarted`, `backgroundTaskComplete`, `backgroundTaskFailed` on `AgentEventBus` so the modal live-updates while open; unsubscribes on close
- `src/services/folder-initializer.ts`: removed `background-tasks` subfolder — output paths now resolve outside the plugin state folder, which validators reject
- `prompts/agentRulesPrompt.hbs`: new **BACKGROUND TASKS** section — when to use `background: true`, always provide `output_file`/`output_path`, retrieve results with `read_file`
- Captured service references (`deepResearch`, `imageGeneration`) before async callbacks to avoid stale closures
- 18 new unit tests across `test/tools/deep-research-tool.test.ts` and `test/tools/image-tools.test.ts` including callback-invocation tests that pull and invoke the submitted callback with `isCancelled = () => false`
- Updated `test/services/folder-initializer.test.ts` to reflect removal of `background-tasks` subfolder

## Non-goals (intentionally not implemented)

- Session-resume completion banners — status bar Notice + output file are sufficient
- Auto-notifying the agent when a background task finishes — fire-and-forget only
- "Clear recent" button in `BackgroundTasksModal` — deferred as follow-up
- `target_note` as folder hint for research output path — deferred as follow-up

## Screenshots / Screencast

No UI changes — backend/internal only.

### Required

- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — PR #634 is purely backend logic (BackgroundTaskManager, tool callbacks, FolderInitializer). No platform-specific APIs used; all Obsidian vault/fs calls work identically on Mobile.

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Deep research and image generation tools now support background task execution with immediate task ID returns for asynchronous processing.
  * Background tasks modal now live-updates in real-time as tasks start, complete, or fail.

* **Bug Fixes**
  * Status bar now correctly remains visible when RAG is idle with no running tasks.

### DIFF
--- a/prompts/agentRulesPrompt.hbs
+++ b/prompts/agentRulesPrompt.hbs
@@ -19,6 +19,18 @@ For complex tasks involving multiple tool calls, plan before executing.
 - For tasks involving 4+ tool calls, briefly explain your plan to the user first
 - Read files BEFORE modifying or deleting them
 
+**BACKGROUND TASKS** (`deep_research` and `generate_image`):
+Both tools accept an optional `background: true` parameter. Use it when:
+- The operation is one step in a larger plan and you want to continue other work without waiting
+- The user explicitly asks to "run in the background" or "don't wait for it"
+- You have other tool calls to make that don't depend on the result
+
+When using `background: true`:
+- Always provide `output_file` (deep_research) or `output_path` (generate_image) so you know exactly where to find the result
+- The tool returns `{ taskId, output_file }` or `{ taskId, output_path }` immediately
+- Retrieve the result later using `read_file` on the returned path — no new tool call to the original tool is needed
+- Do NOT use `background: true` if subsequent steps immediately depend on the output — use foreground mode instead so the result is available inline
+
 **RESEARCH AND INVESTIGATION**:
 When the user asks questions about their notes, be THOROUGH in your research:
 
--- a/src/services/background-status-bar.ts
+++ b/src/services/background-status-bar.ts
@@ -58,8 +58,9 @@ export class BackgroundStatusBar {
 		const runningCount = this.taskManager.runningCount;
 		const ragStatus = this.ragProvider?.getStatus() ?? 'disabled';
 
-		// Nothing to show — hide entirely
-		if (runningCount === 0 && (ragStatus === 'disabled' || ragStatus === 'idle')) {
+		// Nothing to show — hide only when RAG is fully disabled and no tasks are running.
+		// When RAG is idle, show the database icon + indexed-file count.
+		if (runningCount === 0 && ragStatus === 'disabled') {
 			this.statusBarItem.style.display = 'none';
 			return;
 		}
--- a/src/services/folder-initializer.ts
+++ b/src/services/folder-initializer.ts
@@ -9,7 +9,7 @@ import { ensureFolderExists } from '../utils/file-utils';
  */
 export class FolderInitializer {
 	// Subfolder names relative to the plugin state root
-	private static readonly SUBFOLDERS = ['Agent-Sessions', 'Prompts', 'Skills', 'background-tasks'];
+	private static readonly SUBFOLDERS = ['Agent-Sessions', 'Prompts', 'Skills'];
 
 	constructor(private plugin: ObsidianGemini) {}
 
--- a/src/services/folder-initializer.ts
+++ b/src/services/folder-initializer.ts
@@ -9,7 +9,7 @@ import { ensureFolderExists } from '../utils/file-utils';
  */
 export class FolderInitializer {
 	// Subfolder names relative to the plugin state root
-	private static readonly SUBFOLDERS = ['Agent-Sessions', 'Prompts', 'Skills'];
+	private static readonly SUBFOLDERS = ['Agent-Sessions', 'Prompts', 'Skills', 'background-tasks'];
 
 	constructor(private plugin: ObsidianGemini) {}
 
--- a/src/tools/deep-research-tool.ts
+++ b/src/tools/deep-research-tool.ts
@@ -1,3 +1,4 @@
+import { normalizePath } from 'obsidian';
 import { Tool, ToolResult, ToolExecutionContext } from './types';
 import { ToolCategory } from '../types/agent';
 import { ToolClassification } from '../types/tool-policy';
@@ -111,7 +112,7 @@ export class DeepResearchTool implements Tool {
 				// If the caller didn't specify one, generate a timestamped path inside the
 				// plugin state folder's background-tasks directory.
 				const resolvedOutputFile =
-					outputFile ?? `${plugin.settings.historyFolder}/background-tasks/research-${Date.now()}.md`;
+					outputFile ?? normalizePath(`${plugin.settings.historyFolder}/background-tasks/research-${Date.now()}.md`);
 
 				const deepResearch = plugin.deepResearch;
 				const label = params.topic.length > 40 ? params.topic.slice(0, 37) + '…' : params.topic;
@@ -122,9 +123,6 @@ export class DeepResearchTool implements Tool {
 						scope: params.scope,
 						outputFile: resolvedOutputFile,
 					});
-					if (context.session && result.outputFile) {
-						context.session.context.contextFiles.push(result.outputFile);
-					}
 					return result.outputFile?.path;
 				});
 
--- a/src/tools/deep-research-tool.ts
+++ b/src/tools/deep-research-tool.ts
@@ -124,7 +124,7 @@ export class DeepResearchTool implements Tool {
 					// Ensure the parent folder exists before conductResearch tries to save there.
 					const folder = resolvedOutputFile.includes('/') ? resolvedOutputFile.split('/').slice(0, -1).join('/') : null;
 					if (folder) {
-						await ensureFolderExists(plugin.app.vault, folder, folder, plugin.logger);
+						await ensureFolderExists(plugin.app.vault, folder, 'output directory', plugin.logger);
 					}
 
 					// Poll for cancellation every 2 s and signal the API if the task is cancelled.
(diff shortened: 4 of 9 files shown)


allenhutchison:
# Review

Thanks for the work on this — the scaffolding is clean and the tests are substantive.

**A note on testing before review.** Before getting into specifics, I want to gently flag expectations: I expect contributors to manually exercise the happy paths on their changes before requesting review. When I tested this PR, neither tool worked end-to-end with the documented auto-generated path pattern — bug (1) below means every background call without an explicit output path throws. The `BackgroundTasksModal` also doesn't live-update, and cancellation doesn't actually stop work. These would have surfaced immediately with a basic "submit a background research task, watch the modal, try to cancel it" smoke test.

The PR checklist includes "I have tested this change on Desktop" — please treat that as a substantive check, not a box to tick. Jest coverage is great, but it's not a substitute for actually running the feature. Given this PR is specifically about user-facing async UX (status bar, modal, completion notices), manual testing is especially load-bearing here.

I'll still do thorough reviews, but catching happy-path breakage before I see it will make the review cycle faster for both of us. Specifics below.

## 🔴 Blocking bugs (manual testing)

### 1. Background auto-path fails for both tools
The PR falls back to `{historyFolder}/background-tasks/` when no explicit output path is given, but both underlying services **explicitly reject paths inside the plugin state folder**:
- `src/services/image-generation.ts:127-134` — `validateOutputPath` throws "Output path cannot be inside the plugin state folder"
- `src/services/deep-research.ts:298-308` — same check for research reports

Every background call without an explicit path will fail when the callback runs. The PR description's claim that `FolderInitializer` makes the path "always writable" is incorrect — the folder is created, but the validators refuse to write there. See filename/location recommendations under (7) below.

### 2. Cancellation doesn't actually cancel
- `deep-research-tool.ts:119-127` and `image-tools.ts:98-101` check `isCancelled()` once at the top of the callback, then `await` the full API call.
- Result: cancelling a running task marks it cancelled in the manager, but the research/image work completes and the output file is still written to disk.
- `DeepResearchService.cancelResearch()` already exists (`deep-research.ts:207`) and isn't wired up. Image generation would need a similar hook (AbortController or equivalent).

### 3. `BackgroundStatusBar` hides RAG idle state (regression)
- Early return at `background-status-bar.ts:62` hides the widget when `runningCount === 0 && ragStatus === 'idle'`.
- The idle-rendering branch at lines 80-92 is unreachable.
- Regression vs. the old `RagStatusBar`, which showed the database icon + indexed-file count at idle. This contradicts #589's decision to use the status bar as "the canonical 'things are happening in the background' surface."

### 4. `BackgroundTasksModal` doesn't live-update
Rendered once in `onOpen()` with no event subscriptions. Tasks that complete or fail while the modal is open stay frozen as "running". Needs to subscribe to `backgroundTaskStarted`/`Complete`/`Failed` on `AgentEventBus` and re-render (or register on the manager directly).

## 🟡 Correctness / UX

### 5. Image path should route through the attachments folder
Foreground `generateImage` (no explicit `output_path`) always resolves via `app.fileManager.getAvailablePathForAttachment(filename, referenceNotePath)` (`image-generation.ts:196-208`), which respects the vault's configured attachment-folder setting. Background mode should do the same — dumping images into `historyFolder/background-tasks/` bypasses user configuration and surprises users.

Capture the active file path **at submit time** (before the callback runs), since the active file can change while the task is in flight.

### 6. Research should honor `target_note` as a folder hint
No pre-existing default-filename convention for research, so this PR is defining one. Recommend: if the caller provided `target_note`, resolve relative to that note's folder; only fall back to `{historyFolder}/background-tasks/` as a last resort (and only after fixing the validator issue in bug #1).

### 7. Filenames should follow the project's `YYYY-MM-DD <Title>.ext` convention
Agent sessions use `YYYY-MM-DD <Title>.md` (`session-manager.ts:73`, `agent-view-session.ts:251-253`) via `formatLocalDate()` (local timezone, not UTC) + `sanitizeFileName()` — see `src/utils/format-utils.ts:10` and `src/utils/file-utils.ts:123`.

Background-generated filenames should match this pattern for UX consistency:
- Research: `YYYY-MM-DD <sanitized-topic>.md` (replace `research-${Date.now()}.md`)
- Image: `YYYY-MM-DD <sanitized-prompt>.png` (replace `image-${Date.now()}.png`; also replaces the foreground `generated-...-${timestamp}.png` pattern in `image-generation.ts:187-194` for consistency — that pattern predates this convention)

Reuse the existing `formatLocalDate()` + `sanitizeFileName()` helpers instead of re-rolling. Uniqueness is handled by `getAvailablePathForAttachment` (images) or an equivalent numeric suffix (research) — raw `Date.now()` in the filename isn't necessary.

## 🟢 Minor

### 8. Subfolder naming inconsistency
`folder-initializer.ts:12` — existing subfolders are PascalCase (`Agent-Sessions`, `Prompts`, `Skills`); new one is lowercase `background-tasks`. Rename to `Background-Tasks`. Also update the folder-structure list in `AGENTS.md`.

### 9. PR description disagrees with the code
> `src/tools/image-tools.ts`: … returns `output_path: null` when not provided

Code actually auto-generates a timestamped path and returns it — the implemented behavior is better. Please update the PR body so future archaeology matches reality.

### 10. Tests: exercise the submitted callback body
The background-mode tests assert `BackgroundTaskManager.submit` was called with *a* function but never invoke it. One additional test per tool that pulls the captured callback (`mock.calls[0][2]`), invokes it with `isCancelled = () => false`, and asserts the service was called with the expected args would catch future drift between the foreground and background paths. Given the high coverage bar on this project, worth adding. (Notably, this kind of test would have caught bug #1.)

### 11. Recent tasks have no dismiss/clear affordance
Only auto-pruned past `MAX_RECENT = 20`. Consider a "Clear recent" button in `BackgroundTasksModal` + a `clearRecent()` method on the manager.

## Out of scope (noted, not requested here)

- **`contextFiles` registration on background-task completion** — deferred per #589's fire-and-forget decision; agent retrieves via `read_file` (already documented in `agentRulesPrompt.hbs`).
- **Command palette "Generate Image" routing through the background system** — tracked separately as #652.

## What works

Agent background flow, status bar click-to-modal, and completion Notice "Open result" link all work correctly when an explicit output path is provided.

saheersk:
Hey @allenhutchison, thanks for the thorough review — all blocking issues addressed. Here's what was fixed and tested:

**Manual testing (addressing review feedback)**

Tested on Desktop (macOS) after addressing all blocking issues:
- Background research with no explicit path → file created at `Background Research/2026-04-19 Byzantine Empire.md` ✅
- Cancel during research → task marked `cancelled`, no file written ✅
- Status bar shows database icon + file count when RAG idle, no tasks running ✅
- `BackgroundTasksModal` live-updates while open  task moved from Running → Recent on completion without closing modal ✅
- Explicit output path end-to-end happy path ✅

**Known issue (pre-existing, out of scope)**

Spotted `DeepResearchService.saveReport` throwing `"File already exists"` during testing `getAbstractFileByPath` returns null on a vault cache miss after a folder is freshly created, so the code falls through to `vault.create` even though the file exists on disk. Pre-existing in the service, not introduced here. The fix is a one-liner fall back to `vault.adapter.exists()` before calling `vault.create`. Happy to include that in a follow-up PR or fold it in here if you'd prefer to keep it contained.

allenhutchison:
Manual testing went well — the four blocking bugs from my prior review are all resolved. Status bar correctly reverts to RAG state when no tasks are running, cancel actually halts research, the modal live-updates while open, and auto-paths land in writable locations.

Going to squash-and-merge so the diff against `master` lands cleanly without you needing to rebase.

**Filed as follow-ups (not requested on this PR):**
- #656 — unified activity modal with Background Tasks + RAG tabs
- #657 — canonical Background-Tasks output folder, revisit after the broader epic lands
- #658 — bias long-running tools toward background by default
- #659 — surface resolved `output_path` when `generate_image` is called with `background: true` and no explicit path

Thanks for the iteration on this one — the cancel poller wire-up and the live-update subscriptions in particular are clean.

PR #673, opened 2026-04-21
test: regression test for recall_sessions matching deleted-file refs

### LINKED ISSUE #506: fix: recall_sessions should match deleted files via raw wikilinks
## Summary

When restoring `accessedFiles` from frontmatter, wikilinks that can't be resolved (deleted files) are silently dropped. This means `recall_sessions` won't find sessions that touched files that have since been deleted, even though the wikilink `[[Deleted Note]]` is still in the session frontmatter.

## Proposed Fix

In the recall tool's file matching logic, fall back to raw string matching against the frontmatter `accessed_files` wikilinks when the resolved `accessedFiles` Set doesn't contain a match. This requires reading the raw frontmatter rather than relying solely on the hydrated Set.

## Related
- #411 — recall_sessions tool
- `src/agent/session-manager.ts` — accessed_files restoration (lines 266-278)

allenhutchison:
## Summary

Fixes #506. The bug described in #506 was incidentally resolved by #583 (2026-04-08, six days after #506 was filed). That PR switched `recall_sessions` away from the hydrated `accessedFiles: Set<TFile>` to a raw-refs path via `sessionManager.getSessionMetadata()` + `extractRawRefs()`. `extractRawRefs` only strips `[[ ]]` — no metadata-cache resolution — so a wikilink to a since-deleted file survives intact in `accessedFileRefs` and still matches via substring.

This PR is a **regression test only**. It makes the "deleted file still matches" scenario explicit in `session-recall-tool.test.ts` so nobody silently reintroduces hydration on the recall path. (The session-manager layer already has a partner test at `test/agent/session-manager.test.ts:612` asserting `getFirstLinkpathDest` is never called.)

No production code changes.

## Changes

- `test/tools/session-recall-tool.test.ts`: new test "still matches sessions whose accessed_files ref a since-deleted file" with a comment pointing to #506 / #583.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#506)
- [x] I have tested this change on Desktop — 1383/1383 tests pass
- [x] I have verified this change does not break Mobile — test-only change, no runtime code touched
- [x] Documentation has been updated — N/A (test-only)

### AI-Generated Code

## Summary by CodeRabbit

* **Tests**
  * Added a regression test to ensure the session recall feature reliably finds sessions referenced by a deleted note’s raw reference string, and only returns matching sessions.

### DIFF
--- a/test/tools/session-recall-tool.test.ts
+++ b/test/tools/session-recall-tool.test.ts
@@ -130,6 +130,31 @@ describe('RecallSessionsTool', () => {
 		expect(titles).toEqual(['Has context']);
 	});
 
+	// Regression for #506: recall uses raw refs from frontmatter (no metadata-cache
+	// resolution), so a ref to a since-deleted file survives in accessedFileRefs and
+	// still matches. If someone reintroduces hydration here, this test will fail
+	// because the deleted file's ref would be dropped at the session-manager layer.
+	it('still matches sessions whose accessed_files ref a since-deleted file', async () => {
+		const withDeleted = makeSession({
+			id: 'a',
+			title: 'Touched a note that was later deleted',
+			accessedFileRefs: ['Deleted Note'],
+		});
+		const unrelated = makeSession({
+			id: 'b',
+			title: 'Unrelated',
+			accessedFileRefs: ['Other'],
+		});
+		const ctx = makeContext({
+			sessionManager: { getSessionMetadata: jest.fn().mockResolvedValue([withDeleted, unrelated]) },
+		});
+
+		const result = await getTool().execute({ filePath: 'Deleted Note' }, ctx);
+		expect(result.success).toBe(true);
+		const titles = (result.data as any).sessions.map((s: any) => s.title);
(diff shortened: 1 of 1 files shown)


allenhutchison:
@coderabbitai — thanks, you're right. The comment's claim that this test catches hydration regressions at the SessionManager layer was wrong: `getSessionMetadata` is stubbed here, so that layer isn't exercised at all.

I went with your option (a) — narrowed the wording. In 058f3f1:

- Title is now "filter still matches a ref string even when its underlying file no longer exists", which is accurate to what the stubbed setup proves.
- Comment calls out that this is the recall-layer half of the regression and points at the partner invariant already in `test/agent/session-manager.test.ts` ("should not call getFirstLinkpathDest (no TFile resolution)"), which covers the SessionManager side.

Didn't add a new SessionManager test because that invariant is already directly asserted — adding a duplicate "deleted file scenario" test there wouldn't cover any code path the existing one misses (neither test actually touches the vault; both prove `getFirstLinkpathDest` is never consulted).

PR #695, opened 2026-04-25
ix: prevent ScheduledTaskManager double-init on plugin reload (#678)

### LINKED ISSUE #678: ScheduledTaskManager: initialize() + start() fires twice on plugin reload
## Background

After the scheduled-task engine landed in #650, `ScheduledTaskManager.initialize()` and `start()` run twice when the plugin is reloaded against an already-ready workspace:

1. `LifecycleService.setup()` — when `layoutReady` is already true (e.g. on `plugin:reload` or a settings-save reinit), the tail of `setup()` calls `initialize() + start()` (`src/services/lifecycle-service.ts:81-95`).
2. `LifecycleService.onLayoutReady()` — fires separately and also calls `initialize() + start()` (`src/services/lifecycle-service.ts:155-158`).

Both are idempotent (listeners are unregistered before re-registering; `start()` guards on `tickIntervalId !== null`), so there is no correctness issue — but you get duplicate log lines, a redundant `getMarkdownFiles()` scan, and a second `saveState()` write per reload.

Observed on `obsidian plugin:reload id=gemini-scribe`:

```
[ScheduledTaskManager] Initialized with 0 task(s)
[ScheduledTaskManager] Tick loop started
[ScheduledTaskManager] Initialized with 0 task(s)
```

## Proposal

Either gate the `setup()`-tail branch on "this is actually a re-init after first layout-ready" (not first boot), or have `onLayoutReady()` no-op when `initialized === true`. The second option is cheaper to reason about — `initialize()` would early-return if already initialized unless an explicit `refresh: true` flag is passed from the settings-change path.

## Acceptance

- [ ] Fresh plugin load runs `initialize()` exactly once
- [ ] `saveSettings()` re-init (which calls `setup()` with `layoutReady === true`) still refreshes the manager
- [ ] No regression to 40/40 scheduled-task tests

saheersk:
## Summary

Fixes the `ScheduledTaskManager` double-initialisation that occurred on every plugin reload when `layoutReady` was already `true`. Closes #678.

## Before / After Behaviour

**Before (broken):**

On `plugin:reload` (or disable → enable) with Obsidian's workspace already
ready, both `LifecycleService.setup()` and `LifecycleService.onLayoutReady()`
independently called `initialize()` + `start()`, producing:

```
[Gemini Scribe] [ScheduledTaskManager] Initialized with 1 task(s)
[Gemini Scribe] [ScheduledTaskManager] Tick loop started
[Gemini Scribe] [ScheduledTaskManager] Initialized with 1 task(s)   ← duplicate
[Gemini Scribe] [ScheduledTaskManager] Tick loop started             ← duplicate
```

Side effects of the duplicate call:
- A redundant `vault.getMarkdownFiles()` scan on every reload
- A second `saveState()` write against the already-loaded state
- Noisy log output making it harder to diagnose real issues

**After (fixed):**

```
[Gemini Scribe] [ScheduledTaskManager] Destroyed
[Gemini Scribe] [ScheduledTaskManager] Initialized with 1 task(s)   ← once only
[Gemini Scribe] [ScheduledTaskManager] Tick loop started             ← once only
```

The settings-save re-init path still works correctly — changing a setting
such as `historyFolder` triggers a full refresh via `{ refresh: true }`.

## Changes

- `src/services/scheduled-task-manager.ts` — `initialize()` accepts an
  optional `{ refresh?: boolean }` parameter and early-returns when already
  initialised unless `refresh: true` is explicitly passed
- `src/services/lifecycle-service.ts` — the settings-change branch in
  `setup()` now passes `{ refresh: true }` to `initialize()` so that
  historyFolder renames and other setting changes are still picked up without
  a full plugin restart; `onLayoutReady()` continues to call `initialize()`
  without the flag, making it a no-op after the first run
- `test/services/scheduled-task-manager.test.ts` — two new regression tests:
  one verifying the no-op behaviour on a plain second call, one verifying that
  `{ refresh: true }` forces a re-run

## Screenshots / Screencast

### Required

### AI-Generated Code

## Release Notes

* **Bug Fixes**
  * Enhanced app initialization by preventing unnecessary task manager reinitialization during startup and lifecycle transitions, improving overall performance.
  * Folder configuration changes are now properly applied immediately when saved, without requiring the application to restart.

### DIFF
--- a/src/services/lifecycle-service.ts
+++ b/src/services/lifecycle-service.ts
@@ -90,7 +90,7 @@ export class LifecycleService {
 					}
 				}
 			}
-			await plugin.scheduledTaskManager.initialize();
+			await plugin.scheduledTaskManager.initialize({ refresh: true });
 			plugin.scheduledTaskManager.start();
 		}
 	}
--- a/src/services/scheduled-task-manager.ts
+++ b/src/services/scheduled-task-manager.ts
@@ -162,10 +162,19 @@ export class ScheduledTaskManager {
 
 	/**
 	 * Discover task definition files and load the sidecar state.
-	 * Idempotent — safe to call again after a settings change (e.g. historyFolder
-	 * rename). Re-registers the metadata cache listener and re-discovers tasks.
+	 *
+	 * On a fresh plugin load this is called once from onLayoutReady().
+	 * On a settings-save re-init it is called from LifecycleService.setup()
+	 * with refresh: true so that historyFolder changes are picked up without
+	 * requiring a full plugin restart.
+	 *
+	 * Passing no arguments (or refresh: false) after the first successful
+	 * initialization is a no-op — this prevents the double-init that occurs
+	 * when setup() runs with layoutReady === true and onLayoutReady() fires
+	 * immediately afterwards.
 	 */
-	async initialize(): Promise<void> {
+	async initialize(options?: { refresh?: boolean }): Promise<void> {
+		if (this.initialized && !options?.refresh) return;
 		// Unregister previous listeners before re-registering so settings
 		// changes (e.g. historyFolder rename) don't leave stale handlers active.
 		if (this.metadataCacheHandler) {
--- a/test/services/scheduled-task-manager.test.ts
+++ b/test/services/scheduled-task-manager.test.ts
@@ -136,6 +136,30 @@ describe('ScheduledTaskManager', () => {
 		});
 	});
 
+	// ── Double-init guard ───────────────────────────────────────────────────
+
+	describe('initialize() idempotency', () => {
+		it('runs only once when called twice without refresh flag (plugin:reload path)', async () => {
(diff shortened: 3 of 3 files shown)


saheersk:
@coderabbitai review

allenhutchison (approved):
Thanks for the fast, focused fix. Verified live on a running Obsidian instance via `plugin:reload`:

```
[ScheduledTaskManager] Destroyed
[ScheduledTaskManager] Initialized with 0 task(s)   ← once
[ScheduledTaskManager] Tick loop started             ← once
```

Settings-save re-init via `lifecycle.setup()` still triggers a fresh `Initialized` log, confirming the `{ refresh: true }` path works. No errors. 50 unit tests pass; types and prettier clean.

Tiny nit (don't fix here): PR title has "ix:" instead of "fix:" — feel free to edit before merging if you'd like a clean commit subject. Approving either way.

PR #650, opened 2026-04-18
 feat: scheduled task engine — manager, markdown storage, headless ru…

saheersk:
## Summary

Implements PR 2a of the background/scheduled tasks work. Closes #633.
Parent: #589 · Depends on: #632 (shares background infrastructure)

Adds the core **Scheduled Task Engine** — a markdown-file-per-task storage
model, a 60-second tick scheduler, a headless agent runner, and a view-only
monitor modal. This PR is the foundation that PRs 1b, 2b, and 2c all build on.

---

### New files

**`src/services/scheduled-task-manager.ts`**
- `ScheduledTaskManager` — manages task definitions, runs a 60s `setInterval`
  tick loop (same cadence as ChatTimer), loads sidecar state from
  `scheduled-tasks-state.json`
- `computeNextRunAt(schedule, from)` — pure function supporting `once`,
  `daily`, `weekly`, `interval:Xm`, `interval:Xh`
- Task discovery on `initialize()` via `vault.getMarkdownFiles()` filtered to
  `Scheduled-Tasks/` (excluding `Runs/`)
- Per-task frontmatter parsing: `schedule`, `enabledTools`, `outputPath`,
  `model`, `enabled`, `runIfMissed`
- Auto-pause after 3 consecutive failures; `resetTask()` to clear pause state
- `runNow(slug)` for manual triggering from the UI
- Metadata cache `changed` handler to hot-reload edited task files
- `nextRunAt` advanced immediately before submitting — prevents double-fire
  when background execution takes longer than 60s

**`src/services/scheduled-task-runner.ts`**
- `ScheduledTaskRunner` — runs a single task as a headless agent session
- `HeadlessConfirmationProvider` — auto-approves all tool calls (no UI
  interaction needed)
- Full tool-loop execution up to 20 iterations, mirroring
  `agent-view-tools.ts` history format (user/model/functionResponse parts)
- Reads-before-writes sorting (same pattern as `AgentViewTools`)
- Writes output with frontmatter (`scheduled_task`, `ran_at`) to the resolved
  `outputPath`
- `{slug}` and `{date}` placeholder resolution in output paths
- Cancellation support via `isCancelled()` polling at each iteration

**`src/ui/scheduled-tasks-modal.ts`**
- `ScheduledTasksModal` — view-only monitor (no CRUD — that's PR 2c)
- Shows each task: schedule badge, next run time, last run time, last error
- Status icons: clock / pause-circle / alert-circle
- **Run now** button (disabled when paused or disabled)
[...]

PR #674, opened 2026-04-21
feat: unified activity modal with Background Tasks + RAG tabs (#656)

saheersk:
## Summary

Converts `BackgroundTasksModal` into a unified tabbed modal that surfaces both background task state and RAG indexing state behind the shared status bar icon — restoring the RAG stats affordance that was lost when #632 replaced the separate RAG status bar.

Fixes #656

Also fixes a related bug where `vault.readBinary()` throws on virtual Obsidian files (e.g. built-in tutorial notes), causing noisy upload errors during RAG indexing. Falls back to an `mtime+size` hash so smart sync continues tracking those files without crashing.

## Changes

- **`src/ui/background-tasks-modal.ts`** — Complete rewrite as `BackgroundTasksModal` with two top-level tabs:
  - **Background Tasks** — existing running/recent task list with live-updates via `AgentEventBus` (`backgroundTaskStarted/Complete/Failed`)
  - **RAG** — inner Overview / Files / Failures sub-tabs; searchable file list (click to open note); controls: Sync Now (disabled when no pending changes), Reindex All, Settings; live-updates via `ProgressListener`
- **`src/services/background-status-bar.ts`** — Status bar click now defaults to Background Tasks tab when any task is running, otherwise RAG tab
- **`styles.css`** — New CSS for `.gemini-activity-modal`, `.gemini-activity-tab-bar`, `.gemini-activity-tab`, `.gemini-activity-tab--active`, `.gemini-activity-tab-icon`, `.gemini-activity-content`; modal width 600px; clickable file row cursor
- **`src/services/obsidian-file-adapter.ts`** — `computeHash()` no longer throws when `vault.readBinary()` fails; falls back to `mtime:X-size:Y` proxy hash; removes overly aggressive `< 10 char` skip threshold
- **`src/services/rag-vault-scanner.ts`** — Downgrade external uploader's per-file error logs from `error` to `warn` (failures are already surfaced in the RAG Failures tab)

### Required

- [x] I have read and agree to the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] I have read and agree to the [AI Policy](../AI_POLICY.md)
- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`)
- [x] I have tested this change on Desktop
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards)
- [ ] Documentation has been updated (n/a — no new settings or user-facing commands; existing modal UI only)
[...]

PR #696, opened 2026-04-25
fix: deduplicate task file parse when vault.create and metadataCache.…

### LINKED ISSUE #679: ScheduledTaskManager: duplicate parse when task file is created
## Background

New task files in `<state-folder>/Scheduled-Tasks/` are parsed twice on creation because both handlers added in #650 react to the same event:

1. `vault.on('create')` handler waits 500 ms then calls `parseTaskFile()` (`src/services/scheduled-task-manager.ts:219-244`)
2. `metadataCache.on('changed')` handler fires right after Obsidian indexes the new file's frontmatter and also calls `parseTaskFile()` (`src/services/scheduled-task-manager.ts:187-214`)

Observed logs from a single `app.vault.create()` call:

```
[ScheduledTaskManager] Task "smoke-test-task" reloaded from disk
[ScheduledTaskManager] Task "smoke-test-task" discovered on create
```

No correctness issue (`tasks.set()` is idempotent; the `if (!this.state[slug])` guard prevents duplicate state seeding), but there are two vault reads, two parses, and potentially two `saveState()` writes per new task.

## Proposal

Either:
- **Option A:** Track slugs freshly seeded by the create-handler in a short-lived `recentlyCreated` set; the changed-handler early-returns if the slug was just created (and clears the set after ~2s).
- **Option B:** Drop the `vault.create` handler entirely and rely on `metadataCache.changed`, which also fires for brand-new files once Obsidian indexes them. The regression test added in #650 mocks `vault.on('create', …)` specifically, so this option would require updating that test.

Option B is simpler but gives up the \"discover immediately on create\" semantic if the user creates a file without frontmatter (no changed event fires). Option A keeps both paths but deduplicates the work.

## Acceptance

- [ ] A newly-created task file triggers exactly one parse + one `saveState()`
- [ ] Hot-reloading an existing task file (edit frontmatter) still triggers a re-parse
- [ ] Hot-discovery regression test from #650 still passes

saheersk:
## Summary

Fixes the double-parse that occurs when a new task file is created in
`Scheduled-Tasks/` — both the `vault.on('create')` handler and the
`metadataCache.on('changed')` handler react to the same file, causing
two vault reads, two parses, and potentially two `saveState()` writes
for a single file creation. Closes #679.

## Before / After Behaviour

**Before (broken):**

Creating a new valid task file in `Scheduled-Tasks/` produced two log
lines and two full parse + saveState cycles:

```
[Gemini Scribe] [ScheduledTaskManager] Task "my-task" reloaded from disk    ← metadataCache.changed
[Gemini Scribe] [ScheduledTaskManager] Task "my-task" discovered on create  ← vault.create (500 ms later)
```

**After (fixed):**

Only one parse runs — the `create` handler owns the file and the
`changed` handler skips it:

```
[Gemini Scribe] [ScheduledTaskManager] Task "my-task" discovered on create  ← once only
```

Hot-reload of existing files (editing frontmatter of an already-tracked
task) is unchanged — only `metadataCache.changed` fires in that path,
so the guard has no effect and the re-parse happens as before:

```
[Gemini Scribe] [ScheduledTaskManager] Task "my-task" reloaded from disk
```

## Why Option A (recentlyCreated set) and not Option B (drop vault.create)

The issue proposed two approaches:

**Option B** — remove the `vault.on('create')` handler entirely and rely
solely on `metadataCache.on('changed')`.
- Simpler code — one handler instead of two.
- BUT: `metadataCache.changed` only fires when Obsidian re-indexes
  frontmatter. If a user creates a task file without frontmatter first
  (e.g. creates an empty note and adds the `schedule:` property
  afterwards via the Properties panel), no `changed` event fires on
  creation, so the scheduler would silently miss the file until the
  next plugin reload.
- Also requires deleting the regression test added in #650 that
  specifically covers `vault.on('create')` hot-discovery.

**Option A** — keep both handlers, deduplicate with a `recentlyCreated` set.
- The `create` handler adds the slug to `recentlyCreated` immediately
  (synchronously, before the 500 ms defer).
- The `changed` handler checks the set and returns early if the slug
  is present — no vault read, no parse, no saveState.
- After 500 ms the `create` handler removes the slug from the set and
  runs the single parse.
- "Discover immediately on create" semantic is preserved for all files,
  including those created without frontmatter.
- The #650 regression test passes unchanged.

Option A was chosen because it eliminates the duplicate work without
giving up any existing behaviour or breaking existing tests.

## Changes

- `src/services/scheduled-task-manager.ts` — added `recentlyCreated`
  private Set; `vaultCreateHandler` adds the slug to the set before
  scheduling the 500 ms defer and removes it at the start of the
  timeout callback; `metadataCacheHandler` early-returns if the slug is
  in the set; `destroy()` clears the set on teardown
- `test/services/scheduled-task-manager.test.ts` — two new regression
  tests: one verifying a single parse when both `vault.create` and
  `metadataCache.changed` fire for the same new file; one verifying
  that the `changed`-only hot-reload path for existing files still works

## Screenshots / Screencast

### Required

### AI-Generated Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate parsing when creating new scheduled tasks.
  * Ensures deferred/parsing timers are canceled on re-initialize or teardown to avoid stale work.
  * Stops delayed search renders when switching tabs in background tasks.

* **UI**
  * Buttons in the background tasks modal explicitly set as non-submit to avoid accidental form submissions.

* **Tests**
  * Added tests covering creation, hot-reload ordering, and deferred-cancel behaviors.

### DIFF
--- a/src/services/scheduled-task-manager.ts
+++ b/src/services/scheduled-task-manager.ts
@@ -195,7 +195,6 @@ export class ScheduledTaskManager {
 		this.pendingDefers.clear();
 		this.recentlyCreated.clear();
 
-
 		// Unregister previous listeners before re-registering so settings
 		// changes (e.g. historyFolder rename) don't leave stale handlers active.
 		if (this.metadataCacheHandler) {
--- a/src/services/scheduled-task-manager.ts
+++ b/src/services/scheduled-task-manager.ts
@@ -147,6 +147,12 @@ export class ScheduledTaskManager {
 	 * to avoid double-parsing when both events fire for the same new file.
 	 */
 	private recentlyCreated = new Set<string>();
+	/**
+	 * IDs of in-flight setTimeout calls from vaultCreateHandler defers.
+	 * Tracked so they can be cancelled by initialize() (on re-init) and destroy()
+	 * to prevent stale callbacks from mutating state after teardown.
+	 */
+	private pendingDefers = new Set<ReturnType<typeof setTimeout>>();
 
 	constructor(private plugin: ObsidianGemini) {}
 
@@ -181,6 +187,15 @@ export class ScheduledTaskManager {
 	 */
 	async initialize(options?: { refresh?: boolean }): Promise<void> {
 		if (this.initialized && !options?.refresh) return;
+		// Cancel any 500 ms defers still waiting from a previous initialization so
+		// stale callbacks cannot fire against the freshly-loaded state.
+		for (const id of this.pendingDefers) {
+			clearTimeout(id);
+		}
+		this.pendingDefers.clear();
+		this.recentlyCreated.clear();
+
+
 		// Unregister previous listeners before re-registering so settings
 		// changes (e.g. historyFolder rename) don't leave stale handlers active.
 		if (this.metadataCacheHandler) {
@@ -245,8 +260,14 @@ export class ScheduledTaskManager {
 				// (which fires before our 500 ms defer) skips this file.
 				this.recentlyCreated.add(file.basename);
 				// Defer until the metadata cache has indexed the new file's frontmatter.
-				setTimeout(() => {
+				// Track the timer so initialize() and destroy() can cancel it if they
+				// run before the 500 ms elapses.
+				const timerId = setTimeout(() => {
+					this.pendingDefers.delete(timerId);
 					this.recentlyCreated.delete(file.basename);
+					// Guard: if the manager was destroyed or re-initialized while the
+					// defer was pending, skip the parse — state has been reset.
+					if (!this.initialized) return;
 					this.parseTaskFile(file)
 						.then(async (task) => {
 							if (!task) return;
@@ -261,6 +282,7 @@ export class ScheduledTaskManager {
 							this.plugin.logger.warn(`[ScheduledTaskManager] Failed to parse new task ${file.path}:`, err)
 						);
 				}, 500);
+				this.pendingDefers.add(timerId);
 			}
 		};
 		this.plugin.app.vault.on('create', this.vaultCreateHandler);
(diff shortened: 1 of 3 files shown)


allenhutchison:
The fix itself looks great — `recentlyCreated` cleanly bridges the `vault.create` / `metadataCache.changed` ordering, and the bonus `pendingDefers` cancellation in commit 8136b86 is solid defense-in-depth.

Verified live by rebasing locally: created a valid task file in `Scheduled-Tasks/`, log showed only `Task "test-696" discovered on create` (no duplicate `reloaded from disk`); editing an existing task's frontmatter still re-parses via the changed handler. All 1438 tests pass, types and prettier clean.

**Quick ask before approval:** PR 695 just merged into master, so this branch needs a rebase. The conflict is in `src/services/scheduled-task-manager.ts` around `initialize()` — combining 695's signature with 696's defer-cancellation block. Resolution is to keep both:

```ts
async initialize(options?: { refresh?: boolean }): Promise<void> {
    if (this.initialized && !options?.refresh) return;
    // Cancel any 500 ms defers still waiting from a previous initialization so
    // stale callbacks cannot fire against the freshly-loaded state.
    for (const id of this.pendingDefers) {
        clearTimeout(id);
    }
    this.pendingDefers.clear();
    this.recentlyCreated.clear();

    // Unregister previous listeners ...
```

`test/services/scheduled-task-manager.test.ts` auto-merges cleanly. Once rebased I'll approve.

saheersk:
Hey @allenhutchison, picked up the two cases from #674 in the latest commit:

- `type="button"` on all three `<button>` elements (Clear, Cancel, RAG file-row)
- `ragDebounceTimer` cleared on tab switch

PR #697, opened 2026-04-25
fix: await task drain before ScheduledTaskManager re-init (#680)

### LINKED ISSUE #680: LifecycleService: document cancel-without-await invariant in scheduled-task drain
## Background

In #650, `LifecycleService.setup()` cancels in-flight scheduled-task background jobs before re-initialising the manager (`src/services/lifecycle-service.ts:86-94`):

```ts
if (plugin.backgroundTaskManager) {
    for (const task of plugin.backgroundTaskManager.getActiveTasks()) {
        if (task.type === 'scheduled-task') {
            plugin.backgroundTaskManager.cancel(task.id);
        }
    }
}
await plugin.scheduledTaskManager.initialize();
plugin.scheduledTaskManager.start();
```

`cancel()` only signals cancellation — it does **not** await the task finishing. So when `initialize()` reloads `state` from disk into the refreshed manager, there may still be an older runner mid-flight (e.g. inside a `generateModelResponse` call).

The reason this is safe today:
- `ScheduledTaskRunner.run()` checks `isCancelled()` before writing output and returns `undefined` on cancellation.
- `ScheduledTaskManager.executeTask()` only calls `saveState()` when `outputPath !== undefined`.
- So a cancelled-but-still-running job can never mutate the refreshed state.

This invariant isn't documented anywhere, and it's fragile: if a future change to `executeTask` updates state before checking the return value, the drain would silently become unsafe.

## Proposal

Add a comment above the cancel loop in `lifecycle-service.ts:86` explaining why cancel-without-await is sufficient, and mirror it in `ScheduledTaskManager.executeTask()` at the `if (outputPath !== undefined)` branch so the two sides stay in sync under future refactoring.

Optional stronger fix: add a `drainScheduledTasks(): Promise<void>` to `BackgroundTaskManager` that awaits all `scheduled-task` jobs' completion callbacks, and call it from the lifecycle path. This removes the implicit coupling between runner cancellation semantics and state-save logic.

## Acceptance

- [ ] Invariant documented at both call sites, OR
- [ ] Explicit drain method implemented and used by lifecycle

saheersk:
## Summary

Adds a `drain()` method to `BackgroundTaskManager` and uses it in
`LifecycleService.setup()` to guarantee that all cancelled
`scheduled-task` background jobs have fully settled before
`ScheduledTaskManager.initialize()` reloads state on a settings-save
re-init. Closes #680.

## Changes

- `src/services/background-task-manager.ts` — added `private runPromises`
  map to track each `run()` promise; new public `async drain(type?)` method
  awaits `Promise.allSettled` on in-flight tasks, optionally filtered by type
- `src/services/lifecycle-service.ts` — added
  `await plugin.backgroundTaskManager.drain('scheduled-task')` after the
  cancel loop and before `initialize()` in the settings-save re-init path
- `test/services/background-task-manager.test.ts` — three new regression
  tests: immediate resolution with no tasks, ordering guarantee after
  cancel + drain, type-filtered drain

## Screenshots / Screencast

### Required

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Added an API to explicitly wait for background task completion or cancellation.

* **Bug Fixes**
  * Fixed a race condition during service reinitialization where stale task callbacks could overwrite freshly reloaded state.

* **Tests**
  * Added comprehensive test coverage for task completion waiting functionality.

### DIFF
--- a/src/services/background-task-manager.ts
+++ b/src/services/background-task-manager.ts
@@ -30,6 +30,12 @@ export class BackgroundTaskManager {
 	private eventBus: AgentEventBus;
 	private tasks = new Map<string, BackgroundTask>();
 	private cancellationFlags = new Map<string, boolean>();
+	/**
+	 * Tracks the in-flight Promise returned by run() for each active task.
+	 * Populated in submit(), cleaned up in run()'s finally block.
+	 * Used by drain() to await full settlement after cancellation.
+	 */
+	private runPromises = new Map<string, Promise<void>>();
 	/** Maximum number of completed/failed tasks kept in memory for the monitoring modal. */
 	private static readonly MAX_RECENT = 20;
 	private nextId = 1;
@@ -64,8 +70,10 @@ export class BackgroundTaskManager {
 		this.tasks.set(id, task);
 		this.cancellationFlags.set(id, false);
 
-		// Fire-and-forget — intentionally not awaited by the caller
-		this.run(id, work);
+		// Fire-and-forget for the caller, but the promise is stored in runPromises so
+		// drain() can await full settlement (e.g. after cancel()) without blocking submit().
+		const p = this.run(id, work);
+		this.runPromises.set(id, p);
 
 		return id;
 	}
@@ -195,6 +203,7 @@ export class BackgroundTaskManager {
 
 			new Notice(`Background task failed: ${task.label}\n${task.error}`, 8000);
 		} finally {
+			this.runPromises.delete(id);
 			this.cancellationFlags.delete(id);
 			this.pruneOldTasks();
 			this.notifyStatusChange();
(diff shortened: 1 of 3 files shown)


saheersk:
@coderabbitai review

saheersk:
@coderabbitai review

allenhutchison (approved):
Thanks — clean fix. `drain()` correctly snapshots in-flight `runPromises` and uses `Promise.allSettled`; the `submit()` ordering is safe because `run()` yields at its first `await` before the `runPromises.set()` line runs. Tests cover the key scenarios.

Verified locally that auto-merging master into the branch produces the correct combined `lifecycle-service.ts` (both `drain('scheduled-task')` and `initialize({ refresh: true })` end up present). 1438 tests pass, types clean.

Squash-merging is fine — no rebase needed.

PR #723, opened 2026-04-27
fix: open CatchUpModal directly on mobile when catch-up is pending

### LINKED ISSUE #719: Catch-up approvals are unreachable on mobile (status bar hidden)
## Summary

On mobile, scheduled-task catch-up approvals are unreachable. Discovered while reviewing #701 — saheersk's docs draft claimed mobile auto-opens the modal, but the actual implementation only sets a badge that mobile users can't see.

## Behavior

`LifecycleService.handleCatchUp()` (added in #700) handles the catch-up flow as:

1. Detect missed runs via `ScheduledTaskManager.detectMissedRuns()`
2. **If `autoRunCatchUp: true`** → submit each silently via `runNow()`
3. **If `autoRunCatchUp: false`** → call `reserveForCatchUp(slugs)` and `backgroundStatusBar.setPendingCatchUpCount(n)` to show the `!` badge

The badge is the only entry point to `CatchUpModal`. On mobile, Obsidian hides the status bar entirely, so the user has no way to:
- Discover that there are pending catch-ups at all
- Open `CatchUpModal` to Run / Skip them

The reserved tasks remain in `catchUpPending` and the tick loop skips them, so they never run *and* never advance — they sit indefinitely.

## Repro

1. On a mobile vault, create a scheduled task with `runIfMissed: true` and a past `nextRunAt` (e.g. via desktop or by editing `scheduled-tasks-state.json` directly)
2. Open the vault on mobile with `autoRunCatchUp: false` (the default)
3. Observe: no UI surface for the pending catch-up; tasks never run

## Affected files

- \`src/services/lifecycle-service.ts\` — \`handleCatchUp()\` (added in #700)
- \`src/services/background-status-bar.ts\` — mobile-hidden status bar is the only badge surface

## Possible fixes

A few options, roughly ordered by simplicity:

1. **Auto-open the modal on mobile only.** Detect mobile (`Platform.isMobile`) inside `handleCatchUp()` and, when not auto-running, open `CatchUpModal` directly instead of just setting the badge. The modal is already designed to be dismissible without acting, and the badge state remains so it works as expected on desktop. Smallest change.

2. **Auto-open the modal universally** when there are catch-ups. Removes the badge-first UX for desktop too — probably too invasive for users who liked the unobtrusive desktop flow.

3. **Default `autoRunCatchUp: true` on mobile.** Sidesteps the unreachability but takes the choice away from the user. Discoverability is still a problem (they don't know catch-ups happened).

4. **Make catch-ups visible in some other surface on mobile** — e.g. a Notice on startup, a command palette entry like \"Review missed scheduled runs\", or a settings panel section. Higher implementation cost.

(1) seems like the right tradeoff: minimal code, preserves the desktop flow, and matches what saheersk's PR #701 docs draft actually claimed the behavior was.

## Related

- #700 introduced \`handleCatchUp()\` and the badge surface
- #701 added the scheduler management UI; review of that PR surfaced the docs/behavior mismatch that exposed this gap

saheersk:
PR Description — fix/mobile-catchup-modal
==========================================

Title: fix: open CatchUpModal directly on mobile when catch-up is pending

## Summary

On mobile, Obsidian hides the status bar entirely, making the `!` badge added by
`BackgroundStatusBar.setPendingCatchUpCount()` unreachable. Tasks reserved for
catch-up via `reserveForCatchUp()` sit indefinitely — the user has no way to
discover or act on them.

Detects `Platform.isMobile` inside `LifecycleService.handleCatchUp()` and opens
`CatchUpModal` directly on startup when missed scheduled runs are pending, instead
of relying solely on the badge. The badge is still set so the existing desktop
flow is completely unchanged.

Fixes #719

## Changes

- `src/services/lifecycle-service.ts`: import `Platform` from obsidian; add mobile
  guard in `handleCatchUp()` — when `Platform.isMobile` is true and
  `autoRunCatchUp` is false, open `CatchUpModal` directly in addition to setting
  the badge
- `test/services/lifecycle-service.test.ts`: add `vi.hoisted()` block for
  `mockPlatform` and `MockCatchUpModalClass`; add `Platform` to the obsidian mock;
  mock `../../src/ui/catch-up-modal`; add two new tests:
    - "opens CatchUpModal directly on mobile when autoRunCatchUp is false"
    - "does not open CatchUpModal on desktop when autoRunCatchUp is false"

## Screenshots / Screencast

No UI changes on desktop. On mobile: `CatchUpModal` now appears automatically on
startup when missed scheduled runs are pending. Previously nothing appeared and
tasks were stuck indefinitely.

Screenshot 2026-04-27 at 10 28 16 PM

### Required

- [x] All CI checks pass (npm test, npm run build, npm run format-check)
- [x] I have verified this change does not break Mobile (fix is gated behind
      Platform.isMobile; desktop path is unaffected)
- [x] Documentation has been updated (if applicable) — no user-facing docs affected
      by this internal platform guard
- [x] I understand that I must address all review comments from CodeRabbit and
      maintainers, or this PR may be closed

### AI-Generated Code

- [x] AI tool(s) used: Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missed scheduled tasks on mobile devices. Users now see an interactive approval modal for pending catch-ups when auto-catch-up is disabled, ensuring they can take action even when status bar indicators aren't visible.

### DIFF
--- a/src/services/lifecycle-service.ts
+++ b/src/services/lifecycle-service.ts
@@ -1,4 +1,4 @@
-import { TFile } from 'obsidian';
+import { TFile, Platform } from 'obsidian';
 import { Notice } from 'obsidian';
 import type ObsidianGemini from '../main';
 import { AgentEventBus } from '../agent/agent-event-bus';
@@ -540,6 +540,13 @@ export class LifecycleService {
 			plugin.scheduledTaskManager.reserveForCatchUp(pending.map((e) => e.task.slug));
 			// Show the ! badge — clicking it opens the CatchUpModal
 			plugin.backgroundStatusBar?.setPendingCatchUpCount(pending.length);
+
+			// On mobile the status bar is hidden, so the badge is unreachable.
+			// Open the approval modal directly so the user can act on missed runs.
+			if (Platform.isMobile) {
+				const { CatchUpModal } = await import('../ui/catch-up-modal');
+				new CatchUpModal(plugin.app, plugin, pending).open();
+			}
 		}
 	}
 
--- a/test/services/lifecycle-service.test.ts
+++ b/test/services/lifecycle-service.test.ts
@@ -1,10 +1,26 @@
 import type { Mock } from 'vitest';
 
+const { mockPlatform, mockCatchUpModal, MockCatchUpModalClass } = vi.hoisted(() => {
+	const mockCatchUpModal = { open: vi.fn() };
+	// Must use a regular function (not arrow) so it can be called with `new`
+	function MockCatchUpModalClass() {
+		return mockCatchUpModal;
+	}
+	return {
+		mockPlatform: { isMobile: false },
+		mockCatchUpModal,
+		MockCatchUpModalClass: vi.fn().mockImplementation(MockCatchUpModalClass),
+	};
+});
+
 vi.mock('obsidian', () => ({
 	TFile: class TFile {},
 	Notice: vi.fn(),
 	normalizePath: (p: string) => p,
+	Platform: mockPlatform,
 }));
+
+vi.mock('../../src/ui/catch-up-modal', () => ({ CatchUpModal: MockCatchUpModalClass }));
 
 vi.mock('../../src/services/tool-registrar', () => ({
 	ToolRegistrar: vi.fn().mockImplementation(function () {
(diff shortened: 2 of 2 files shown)


allenhutchison (approved):
Approving — clean, minimal, well-tested fix that matches the maintainer-recommended approach in #719. Will follow up with a small cleanup pass for nits (consolidating the duplicated `obsidian` import line and a one-line comment clarifying that the badge is intentionally still set on mobile).

PR #724, opened 2026-04-27
feat: consolidate background outputs into [state-folder]/Background-T…

### LINKED ISSUE #657: Follow-up: canonical Background-Tasks output folder
## Background

PR #651 (closes #634) initially provisioned a \`Background-Tasks/\` subfolder under the plugin state folder via \`FolderInitializer\`, but removed it because the deep-research and image-generation services' validators (\`deep-research.ts:298-308\`, \`image-generation.ts:127-134\`) explicitly reject any output path inside the plugin state folder.

Current fallback behaviour:

- **Deep research** (no explicit \`outputFile\`) → \`Background Research/YYYY-MM-DD <topic>.md\` at vault root, created lazily via \`ensureFolderExists\` inside the tool callback.
- **Image generation** (no explicit \`output_path\`) → vault's configured attachment folder via \`getAvailablePathForAttachment\`.

So there is no single canonical "where do background outputs land" folder today.

## Why follow up

This is fine for the initial PR but will likely become awkward as the scheduled-tasks epic progresses (#633, #635, #636). When scheduled tasks start producing output files too, users will want one predictable location to browse them — and the current split (Research at vault root, images in attachments) makes that harder.

## Options to evaluate when this comes up again

1. Relax the state-folder validators to allow writes under a dedicated \`[state-folder]/Background-Tasks/\` subfolder (keeps plugin-owned content in one place).
2. Standardise on a top-level \`Background-Tasks/\` folder at vault root for both tools (visible to users, but pollutes the vault root).
3. Keep the current split and document it clearly.

## Acceptance

Not actionable on its own — revisit once the rest of the background/scheduled-tasks epic lands and we can see what real usage patterns look like.

saheersk:
PR Description — fix/background-tasks-output-folder
=====================================================

Title: feat: consolidate background outputs into [state-folder]/Background-Tasks/

## Summary

Previously deep research fell back to a vault-root `Background Research/` folder
and image generation used the Obsidian attachment folder, giving users two
unpredictable locations for background task outputs. As the scheduled-tasks epic
progresses, users need one predictable place to browse all background outputs.

Provisions a `Background-Tasks/` subfolder under the plugin state folder and
routes both tools there by default, while relaxing the state-folder validators to
allow writes into that one subfolder only.

Fixes #657

## Changes

- `src/services/folder-initializer.ts`: add `Background-Tasks/` to `SUBFOLDERS`
  so it is provisioned alongside `Agent-Sessions/`, `Prompts/`, `Skills/`, etc.
- `src/services/deep-research.ts`: relax `validateAndNormalizeFilePath` to allow
  paths under `[state-folder]/Background-Tasks/` while still rejecting all other
  plugin-state subfolders
- `src/services/image-generation.ts`: same validator relaxation in
  `validateOutputPath`; update `resolveDefaultOutputPath` to write into
  `Background-Tasks/` instead of the Obsidian attachment folder; remove the
  now-unused `resolveReferenceNotePath` helper
- `src/tools/deep-research-tool.ts`: update default output path from vault-root
  `Background Research/<date> <topic>.md` to
  `[state-folder]/Background-Tasks/<date> <topic>.md`
- `test/services/folder-initializer.test.ts`: update call-count and
  `arrayContaining` assertions to include the new subfolder
- `test/tools/deep-research-tool.test.ts`: update path regex and mock path to
  match the new `Background-Tasks/` location

## Screenshots / Screencast

No UI changes — this is a backend output-path consolidation. Users will find
both deep research reports and generated images under
`[state-folder]/Background-Tasks/` instead of vault root / attachment folder.

### Required

- [x] All CI checks pass (npm test, npm run build, npm run format-check)
- [x] I have verified this change does not break Mobile (no platform-specific code;
      FolderInitializer runs identically on mobile)
- [x] Documentation has been updated (if applicable) — no user-facing docs reference
      the old output paths
- [x] I understand that I must address all review comments from CodeRabbit and
      maintainers, or this PR may be closed

### AI-Generated Code

- [x] AI tool(s) used: Claude

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated "Background-Tasks" folder for storing background reports and generated images.
* **Bug Fixes**
  * Refined path validation so outputs are allowed in the Background-Tasks subtree while still blocking other restricted locations.
* **Tests**
  * Updated tests to expect and validate the new Background-Tasks folder location.

### DIFF
--- a/src/services/deep-research.ts
+++ b/src/services/deep-research.ts
@@ -295,12 +295,18 @@ export class DeepResearchService {
 			}
 		}
 
-		// Check if path is inside the plugin's history folder (or is the folder itself)
+		// Check if path is inside the plugin's history folder (or is the folder itself).
+		// Background-Tasks/ is the one allowed subfolder — it is the canonical output
+		// location for background and scheduled-task outputs.
 		const historyFolder = this.plugin.settings.historyFolder;
 		if (historyFolder) {
-			// Normalize the history folder to ensure consistent comparison
 			const normalizedHistoryFolder = normalizePath(historyFolder);
-			if (normalizedPath === normalizedHistoryFolder || normalizedPath.startsWith(normalizedHistoryFolder + '/')) {
+			const backgroundTasksFolder = normalizePath(`${normalizedHistoryFolder}/Background-Tasks`);
+			const insideStateFolder =
+				normalizedPath === normalizedHistoryFolder || normalizedPath.startsWith(normalizedHistoryFolder + '/');
+			const insideBackgroundTasks =
+				normalizedPath === backgroundTasksFolder || normalizedPath.startsWith(backgroundTasksFolder + '/');
+			if (insideStateFolder && !insideBackgroundTasks) {
 				throw new Error(
 					`Cannot write report to plugin state folder: "${historyFolder}". Please choose a different output location.`
 				);
--- a/src/services/deep-research.ts
+++ b/src/services/deep-research.ts
@@ -304,8 +304,7 @@ export class DeepResearchService {
 			const backgroundTasksFolder = normalizePath(`${normalizedHistoryFolder}/Background-Tasks`);
 			const insideStateFolder =
 				normalizedPath === normalizedHistoryFolder || normalizedPath.startsWith(normalizedHistoryFolder + '/');
-			const insideBackgroundTasks =
-				normalizedPath === backgroundTasksFolder || normalizedPath.startsWith(backgroundTasksFolder + '/');
+			const insideBackgroundTasks = normalizedPath.startsWith(backgroundTasksFolder + '/');
 			if (insideStateFolder && !insideBackgroundTasks) {
 				throw new Error(
 					`Cannot write report to plugin state folder: "${historyFolder}". Please choose a different output location.`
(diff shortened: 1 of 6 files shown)


saheersk:
@coderabbitai review

allenhutchison (approved):
Approving — validator logic is correct (traced the four interesting cases: allowed Background-Tasks file, rejected sibling-prefix subfolder, rejected bare state folder, allowed file with non-png extension that gets rewritten), `FolderInitializer` change matches the existing list-driven pattern, and removing the `getActiveFile` dependency in image-generation is a real simplification.

Three small follow-ups I'll handle in a separate PR so this one isn't held up further:

1. Update `docs/guide/background-tasks.md` lines 31 and 45 — both still reference the old vault-root `Background Research/` location.
2. Clean up the now-dead `targetNotePath` parameter on `resolveOutputPath` / `resolveDefaultOutputPath` / `saveImageToVault` (and the `image-tools.ts` caller + test that still passes it). The default flow no longer uses an attachment-folder reference, so the parameter is silently dropped.
3. Add unit tests for both validators covering the new allowlist behavior — `[state-folder]/Background-Tasks/foo.md` allowed, `[state-folder]/Skills/foo.md` still rejected, sibling-prefix `[state-folder]/Background-Tasks-Other/foo.md` still rejected. Currently neither validator has direct test coverage.

PR #730, opened 2026-04-29
feat: add custom API endpoint configuration

baozaotumao2025:
**What does this PR do?**
This PR adds a "Custom API Endpoint" text field in the Advanced Settings under API Configuration. This allows users to pass a custom `baseUrl` to the `@google/genai` SDK, which is extremely useful for users utilizing local proxies or API gateways.

**Changes made:**
- Modified `src/main.ts`: Added `customBaseUrl` to `ObsidianGeminiSettings` interface and default settings.
- Modified `src/ui/settings-api.ts`: Added a text input field for the custom endpoint under the Advanced API settings.
- Modified `src/api/gemini-client.ts`: Updated `GoogleGenAI` initialization to utilize `httpOptions.baseUrl` if provided by the user.

**Testing:**
Tested locally in Obsidian. The setting correctly saves, and requests are successfully proxied through the configured base URL without affecting the default behavior when the field is left blank.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable "Custom Base URL" in plugin settings so users can route requests through a proxy or alternate API endpoint.
  * New settings field trims input and persists changes automatically.
  * When set, the plugin sends requests using the configured base URL; default behavior remains unchanged if left empty.

### DIFF
--- a/src/api/gemini-client.ts
+++ b/src/api/gemini-client.ts
@@ -60,11 +60,9 @@ export class GeminiClient implements ModelApi {
 		};
 		this.plugin = plugin;
 		this.prompts = prompts || new GeminiPrompts(plugin);
-		//this.ai = new GoogleGenAI({ apiKey: config.apiKey });
 		const customUrl = this.plugin?.settings?.customBaseUrl;
 		this.ai = new GoogleGenAI({
 			apiKey: config.apiKey,
-			// 将 baseUrl 嵌套在 httpOptions 中
 			httpOptions: customUrl ? { baseUrl: customUrl } : undefined,
 		});
 	}
--- a/src/api/gemini-client.ts
+++ b/src/api/gemini-client.ts
@@ -60,9 +60,14 @@ export class GeminiClient implements ModelApi {
 		};
 		this.plugin = plugin;
 		this.prompts = prompts || new GeminiPrompts(plugin);
-		this.ai = new GoogleGenAI({ apiKey: config.apiKey });
+		//this.ai = new GoogleGenAI({ apiKey: config.apiKey });
+		const customUrl = this.plugin?.settings?.customBaseUrl;
+		this.ai = new GoogleGenAI({
+			apiKey: config.apiKey,
+			// 将 baseUrl 嵌套在 httpOptions 中
+			httpOptions: customUrl ? { baseUrl: customUrl } : undefined,
+		});
 	}
-
 	/**
 	 * Generate a non-streaming response
 	 */
--- a/src/main.ts
+++ b/src/main.ts
@@ -69,6 +69,7 @@ export interface ObsidianGeminiSettings {
 	initialBackoffDelay: number;
 	streamingEnabled: boolean;
 	allowSystemPromptOverride: boolean;
+	customBaseUrl: string;
 	temperature: number;
 	topP: number;
 	stopOnToolError: boolean;
@@ -118,6 +119,7 @@ const DEFAULT_SETTINGS: ObsidianGeminiSettings = {
 	initialBackoffDelay: 1000,
 	streamingEnabled: true,
 	allowSystemPromptOverride: false,
+	customBaseUrl: '',
 	temperature: 0.7,
 	topP: 1,
 	stopOnToolError: true,
--- a/src/ui/settings-api.ts
+++ b/src/ui/settings-api.ts
@@ -62,6 +62,20 @@ export async function renderApiSettings(
 
 	// API Configuration
 	new Setting(containerEl).setName('API Configuration').setHeading();
+	// === 【新增的代理设置面板区 开始】 ===
+	new Setting(containerEl)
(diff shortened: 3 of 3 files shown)


allenhutchison (changes requested):
Thanks for taking the time to put this together — a custom endpoint setting is a useful capability and we'd like to land it. There's some work needed before we can accept it though, both on the code itself and on the contribution process. I want to be upfront about both so the next iteration is straightforward.

## Code

**Blockers**

1. **English-only UI strings.** The setting name, description, and placeholder all contain Chinese text. Every other UI string in the project is English; we can't ship mixed-locale UI. Suggested replacement:
   ```ts
   .setName('Custom API endpoint')
   .setDesc('Override the default Google API base URL (e.g. for a corporate proxy or local gateway). Leave blank to use the official endpoint.')
   .setPlaceholder('https://my-proxy.example.com')
   ```

2. **The setting only routes ~14% of SDK calls.** \`new GoogleGenAI({ apiKey })\` is constructed in **seven** places in this codebase; this PR patches one. The other six (\`tools/web-fetch-tool.ts:74\`, \`:246\`, \`tools/google-search-tool.ts:52\`, \`services/rag-indexing.ts:121\`, \`services/deep-research.ts:61\`, \`services/context-manager.ts:105\`) still hit the public Google endpoint regardless of this setting. For the stated use case (corporate proxy, local gateway), this is worse than not having the feature: in a network that blocks the public endpoint those features will fail mysteriously, and in one that allows both, the user's data is exfiltrated outside the proxy boundary they thought they configured. Either fix all seven call sites — ideally by centralising in a small \`createGoogleGenAI(plugin)\` helper so future call sites stay correct by default — or document the scope very clearly with a warning in the setting description.

3. **Stale begin/end comments.** \`// === 【新增的代理设置面板区 开始】 ===\` and the matching close marker need to come out. Our \`AGENTS.md\` explicitly calls these out as an anti-pattern.

**Should fix before merge**

4. **Bypass of \`GeminiClientConfig\`.** Other config flows through the config object built by \`GeminiClientFactory\`. This PR reaches into \`this.plugin?.settings?.customBaseUrl\` directly inside the constructor, which silently drops the value when callers (notably tests) construct without a plugin reference. Add \`baseUrl?: string\` to \`GeminiClientConfig\` and populate from the factory — mirrors the existing \`OllamaClientConfig.baseUrl\` pattern.

5. **No tests.** A single test verifying that \`GoogleGenAI\` receives \`httpOptions: { baseUrl: '...' }\` when the setting is set, and \`httpOptions: undefined\` when it isn't, would catch the next regression.

6. **No docs update.** A new user-facing setting needs an entry in \`docs/reference/settings.md\` (or \`advanced-settings.md\`).

**Nits** — restore the blank line between the constructor and \`generateModelResponse\` (the \`-\` after the closing brace removed it); consider validating the URL with \`new URL(value)\` so non-URL input fails early with a clear error rather than an obscure SDK one.

## Contribution process

This part is harder to read in a PR comment so I want to be direct: this PR didn't follow the contribution flow we ask of every contributor, and we'll need it to before we can land changes.

- **No PR template.** \`.github/PULL_REQUEST_TEMPLATE.md\` has required sections — \`Fixes #<issue>\`, the **Required** checklist, the **AI-Generated Code** disclosure, screenshots for UI changes — none of which appear in this PR. The template isn't optional; it's how we verify the contributor has read \`CONTRIBUTING.md\` and \`AI_POLICY.md\`.

- **No approved issue first.** \`CONTRIBUTING.md\` asks contributors to open an issue describing the use case, get maintainer alignment on the approach, *then* write the code. Had that conversation happened, the partial-coverage problem (#2 above) would have surfaced before any code was written, saving you the rework.

- **Testing claim doesn't match what shipped.** The description says you tested locally and the proxy works. The proxy works for the chat path you patched; it demonstrably doesn't work for the six other call sites. That's a sign the testing was narrower than the description implies — please be specific in future PRs about exactly which paths you exercised.

- **The repo ships a \`create-pr\` skill** at \`.claude/skills/create-pr/\` that enforces the template and runs the pre-flight checks (\`npm run format-check\`, \`npm run build\`, \`npm test\`) automatically. We strongly recommend using it; failing that, run those three commands locally and complete the template by hand.

## What I'd suggest

Rather than iterate on this PR, the cleanest path is:

1. Open an issue describing the use case and proposed scope (single setting? per-provider? which call sites?) and wait for a maintainer to confirm the approach.
2. Re-open the PR using the template, with the issue linked, the full checklist completed, the AI-Generated Code section filled in if applicable, plus tests and docs.
3. Run the \`create-pr\` skill (or at minimum its pre-flight checks) before pushing.

Happy to engage on the design discussion in the issue. The feature is welcome — we just need it to land in a way that holds up for the proxy use case it's claiming to solve.

allenhutchison:
Superceeded by #760

Issue #763, opened 2026-05-05
Repeated Agents.md System prompt injection bloats every call


ddavisqa:
I used Scribe this weekend to hold my hand in setting up and completing my first vibe-coding session (a complete refactor of an old python app).  I ran out of tokens after ~45 min.  AI Studio revealed my input was exponentially outpacing my output.  

<img width="598" height="371" alt="Image" src="https://github.com/user-attachments/assets/a21a9705-3fbf-477d-8d97-1ef8276ab3f5" />

<img width="569" height="354" alt="Image" src="https://github.com/user-attachments/assets/f194e72f-f891-4611-a405-6a6d836e4628" />

<img width="580" height="349" alt="Image" src="https://github.com/user-attachments/assets/6712d6b8-5cc2-4505-b873-c056d6664774" />

---

I exported two days worth of session data and ran it through several bots for analysis and they all recommended implementing explicit context caching either in the plugin or in AI Studio, which is not available yet.  

**Prompt**:
Review this code base and tell me how to use caching efficiently https://github.com/allenhutchison/obsidian-gemini

Let me fetch the source files to understand the codebase better.Good, I now have a solid understanding of the codebase. Let me compile actionable caching recommendations based on what I've read.

Here's a thorough breakdown of how to use caching efficiently in this codebase, organized by where the opportunities actually exist:

* * *

Caching Opportunities in `obsidian-gemini`
------------------------------------------

### 1\. Vault Context / AGENTS.md — Avoid Re-reading Every Request

The most expensive redundant I/O in the codebase is the system instruction construction in `src/api/gemini-client.ts`. Every call to `buildGenerateContentParams()` re-reads `plugin.agentsMemory.read()` (the AGENTS.md vault map) and `plugin.skillsService.getSkillSummaries()`. These rarely change mid-session.

**What to do:** Introduce an in-memory cache for the vault context with a dirty flag. Invalidate only when the vault emits a file change event for AGENTS.md:

    
    // In your plugin class or a dedicated VaultContextCache
    private cachedAgentsMemory: string | null = null;
    private agentsMemoryDirty = true;
    
    async getAgentsMemory(): Promise<string> {
      if (!this.agentsMemoryDirty && this.cachedAgentsMemory !== null) {
        return this.cachedAgentsMemory;
      }
      this.cachedAgentsMemory = await this.agentsMemory.read();
      this.agentsMemoryDirty = false;
      return this.cachedAgentsMemory;
    }
    

[...]

Issue #717, opened 2026-04-26
evals: baseline-compare workflow — track regressions across runs


allenhutchison:
## Motivation

\`npm run eval:compare\` exists but requires manually pointing at two result files. To answer the question the harness is *for* (\"is this version better/worse than the last?\"), the workflow needs to be:

1. Run eval
2. \"Did anything regress vs baseline?\"

Today step 2 is: remember which result file is the baseline, type \`eval:compare a.json b.json\`, eyeball the diff. That doesn't scale to running evals routinely.

## Proposal

1. **Named baselines.** \`evals/baselines/\` directory with:
   - \`baselines/gemini-flash.json\` (latest blessed Gemini run)
   - \`baselines/ollama-gemma4-latest.json\` (latest blessed Ollama run)
   - One per (provider, model) pair we care about

2. **\`npm run eval\` auto-compare.** After a run finishes, automatically compare against the matching baseline (\`baselines/<provider>-<model>.json\`) and print a regression summary:

   \`\`\`
   === Regression check vs baseline (2026-04-26) ===
   pass^3: 100% → 100% (=)
   solve^3: 50.0% → 42.9% (-7.1pp)
   Tasks with degraded solve rate:
     find-tagged-notes  3/3 → 0/3
   \`\`\`

3. **\`npm run eval:bless\` to update baselines** — explicit operator action to promote the current run's result file to the baseline. Keeps the baseline intentional, not auto-drift.

4. **CI check (optional)** — fail the build if solve^k regresses by more than X percentage points vs baseline. Out of scope for first pass; mention here so the data structure is forward-compatible.

## Acceptance

- After a fresh eval run, the operator sees a regression summary without typing a separate command
- \`eval:bless\` exists and updates the right baseline file
- Baselines are checked into the repo (small JSON files) so the regression baseline ships with the codebase

## Code refs

- `evals/lib/compare.mjs`
- \`package.json\` scripts

Tracking: #687.

Issue #715, opened 2026-04-26
evals: runtime reliability — per-task timeout, clean interrupt, streaming progress


allenhutchison:
## Motivation

While running PR #694 evals against an Ollama gemma4:31b dense model, three gaps surfaced:

1. **No per-task timeout.** The 31b model was slow enough that a single task could hang for many minutes with no upper bound. Wasted real wall time before realizing the model was just too slow for this hardware.

2. **No streaming progress.** Output is silent between \"Sending message...\" and \"Turn completed.\" During a slow run this is indistinguishable from a hang. Have to attach to obsidian or watch \`ollama ps\` to confirm the harness is alive.

3. **Interrupt cleanup is partial.** The existing handler in `run.mjs` restores \`chatModelName\` if \`--model=\` was used, but doesn't clean up partial fixtures, in-flight sessions, or print a partial-results summary. Killing mid-run leaves the vault with eval-scratch files.

## Proposal

1. **Per-task wall-time budget** — config: `taskTimeoutMs` (default 5 min, override per task). On timeout: cancel the agent loop, mark the task as `timeout` (not pass/fail), continue to next task. Counts as a non-pass for `pass^k`.

2. **Per-turn progress tick** — after each model turn, log: `  [turn N: M tool calls, ETA based on running average]` so the operator can see the eval is alive and roughly when to expect completion.

3. **Cleanup on SIGINT/SIGTERM** — extend existing handler to:
   - Print partial results summary (tasks completed so far, current task in progress)
   - Clean fixture files from the vault
   - Call `agentView.cancelCurrentRun()` on the in-flight session
   - Exit with non-zero so CI / scripts can distinguish interrupted from passing

## Acceptance

- A 1-minute timeout test (\`MOCK_SLEEP=120 npm run eval -- --task=read-and-answer\`) cancels and reports timeout in <70s wall time
- During a normal run, every turn produces a visible progress line within 2s of the model returning
- Ctrl-C during a run leaves no eval-scratch files behind and prints `=== Interrupted: 3 of 7 tasks completed ===`

## Code refs

- `evals/run.mjs` (signal handlers, runTask loop)
- `evals/lib/runner.mjs` or wherever the agent loop is invoked

Tracking: #687.

Issue #713, opened 2026-04-26
evals: matcher brittleness — adopt richer rubric or LLM-as-judge for phrasing-sensitive tasks


allenhutchison:
## Motivation

Two PR #694 eval runs (gemma4:26b and gemma4:latest) both showed the same pattern: `find-tagged-notes` and `multi-file-summary` reported solve rate 0/3 even though the agent did the right thing. The misses are matcher brittleness, not behavior failures.

Concrete examples observed:

- **find-tagged-notes**: agent returned wikilink slugs (\`[[neural-networks]]\`); matcher wanted the title-case string \"Neural Networks\". Both forms are correct in Obsidian context.
- **multi-file-summary**: agent's prose summary used different phrasing than the literal substrings the matcher looks for. Content was right, words were different.

This matters because if the harness is going to be the goto regression detector (per #687), solve rate has to mean \"the agent did the wrong thing\" — not \"the agent phrased it differently.\" Today a model swap or prompt tweak that changes phrasing while preserving behavior gets penalized.

## Proposal

Two paths, probably layered:

1. **Richer rubric language** — extend matchers to accept multiple equivalent forms: \`{ contains: [ \"Neural Networks\", \"[[neural-networks]]\" ] }\`. Cheap, deterministic, no judge call. Catches the find-tagged-notes class.

2. **LLM-as-judge fallback** with a pinned judge model (per #687's reliability methodology) for tasks where output is prose-heavy. Tasks opt in via rubric metadata. Catches the multi-file-summary class.

Per the #687 reliability notes, the judge model must be pinned across model-swap runs so the judgment itself isn't a moving target.

## Acceptance

- find-tagged-notes solves 3/3 on gemma4:latest using one of the two paths
- multi-file-summary solves 3/3 on gemma4:latest under the LLM-judge path
- Pass rate stays stable on Gemini runs (no regression on the path that already works)

## Code refs

- `evals/lib/scorer.mjs` (current rubric application)
- `evals/tasks/find-tagged-notes.json`, `evals/tasks/multi-file-summary.json` (the two affected fixtures)

Tracking: #687.

PR #770, opened 2026-05-07
fix(agent): truncate old tool-result payloads in conversation history

allenhutchison:
## Summary

Tracked under #763. Long agent sessions (especially coding sessions full of `read_file` calls) show input tokens climbing roughly quadratically while output stays flat. The biggest single accumulator is `read_file` returning hundreds of KB of file contents that get appended to history as a `functionResponse` payload, then replayed verbatim on every subsequent send.

This change adds a deterministic, always-on truncation pass: when history is prepared for a request, older tool-result turns whose serialized response exceeds 4 KB get their `response` payload replaced with a small elision marker:

```js
{ success: <original.success>, truncated: true, truncatedFrom: 132489, note: 'Tool result truncated… Re-call the tool if you need the full output.' }
```

The most recent two tool-result turns are left intact so the agent reasoning across a small batch of recent tool calls still has the full text. Smaller responses (≤ 4 KB) pass through unchanged. The pass runs from `ContextManager.prepareHistory`, *before* the token-threshold check, so a session that would have crossed the compaction threshold often stays under it and skips the more expensive summarization round-trip entirely.

Now that #717 (auto-compare baseline), #713 (judge matchers), and #715 (runtime reliability) are landed, the eval harness can measure the impact of this change on solve rate / cost / cache ratio across runs — exactly the signal we built that runway for.

## Changes

- `src/agent/agent-loop-helpers.ts` — new `truncateOldToolResults(history, opts?)`. Pure, side-effect-free, returns a new array. Defaults: 4 KB cap, keep most recent 2 tool-result turns.
- `src/services/context-manager.ts` — `prepareHistory` runs the truncation pass before the threshold check. Logs the byte delta so the operator can see what was shed.
- `docs/reference/settings.md` — new "Tool-result truncation in history" subsection under Context Management.
- `test/agent/agent-loop-helpers.test.ts` — 8 cases covering empty input, no-tool-result histories, keep-recent boundary, mixed sizes, inlineData passthrough (sibling parts must survive), success-flag preservation, custom maxBytes/keepRecent options, and immutability of the input array.

## What this does NOT do

- Doesn't change `buildFunctionResponseParts` — the *fresh* tool result the agent uses on the current turn is still full text. Truncation only affects how that result appears in history on subsequent requests.
- Doesn't deduplicate identical reads (separate idea, deferred).
- Doesn't replace #536 (Files API for binary attachments). That's the cleaner architectural fix for the same class of problem; this PR is the immediate stopgap that needs no API surface changes.
- Doesn't expose a setting yet. Always-on with sensible defaults; we'll add a setting if there's demand.

## Verification

- `npm test` — 1667 tests passing (+8 in `agent-loop-helpers.test.ts`).
- `npm run typecheck:test` — clean.
- `npm run build` — clean.
- Live measurement against the eval harness will happen post-merge with `npm run eval` + `eval:bless` — that's the protocol we wired in #717. Expect lower mean prompt tokens on the multi-turn tasks (`multi-hop-retrieval`, `loop-trap-cyclic-refs`, `create-note-from-search`) without solve-rate regression.

## Screenshots / Screencast

N/A — internal change to history-prep logic.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#763)
- [x] I have tested this change on Desktop — pure-helper logic has unit coverage; live integration impact will be measured via the eval harness as planned.
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — N/A, `agent-loop-helpers.ts` is platform-neutral and runs on both.
- [x] Documentation has been updated — new subsection in `docs/reference/settings.md`.

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Automatic truncation of large tool outputs in conversation history (~4 KB threshold) while preserving the most recent results and allowing full output to be restored by re-running the tool.
  * History preprocessing reduces estimated token counts and can prevent unnecessary compaction.

* **Documentation**
  * Added guidance on tool-result truncation and its interaction with context management.

* **Tests**
  * Added tests validating truncation behavior, recent-result preservation, non-destructive handling, and compaction-skip scenarios.

### DIFF
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -209,6 +209,10 @@ Context management automatically monitors and controls conversation size to prev
 - **How it works**: When conversation tokens exceed this percentage, older turns are summarized and replaced with a compact summary while preserving recent messages
 - **Hard ceiling**: Aggressive compaction triggers at 80% of the input limit to prevent API errors
 
+### Tool-result truncation in history
+
+In addition to compaction, the plugin sheds bloat from older tool-result turns before each request. When a tool (typically `read_file`) returns a payload larger than ~4 KB, the response in conversation history is replaced with a small `{ truncated: true, truncatedFrom: N, note: "..." }` marker once it falls outside the most recent two tool-result turns. This keeps the agent's reasoning intact for the current and immediately previous tool call while preventing the long tail of past tool responses from dominating every subsequent prompt. Re-issuing the original tool call brings the full output back when needed. The behavior is always-on and not currently exposed as a setting.
+
 ### Show Token Usage
 
 - **Setting**: `showTokenUsage`
--- a/src/agent/agent-loop-helpers.ts
+++ b/src/agent/agent-loop-helpers.ts
@@ -177,3 +177,96 @@ export function buildToolHistoryTurns(args: {
 
 	return updated;
 }
+
+/**
+ * Default per-tool-result size cap before we treat a stored response as
+ * bloat worth shedding from history. 4 KB comfortably covers prose answers,
+ * structured JSON, and short file fragments; anything larger is usually a
+ * `read_file` of source code that the model has already digested and won't
+ * need verbatim again.
+ */
+export const DEFAULT_TOOL_RESPONSE_TRUNCATE_BYTES = 4096;
+
+/**
+ * Default number of most-recent tool-result turns to leave intact. Two
+ * gives the agent the just-executed turn plus the previous one (so a model
+ * that's reasoning across a small batch of recent tool calls still has the
+ * full text), while older results — the long tail that drives quadratic
+ * input growth (#763) — get shed.
+ */
+export const DEFAULT_TOOL_RESPONSE_KEEP_RECENT = 2;
+
+/**
+ * Build the elision marker that replaces a `functionResponse.response`
+ * payload when it's truncated. Preserves whatever the original `success`
+ * flag was so loop-detection / scoring code that switches on success keeps
+ * working, and tells the model to re-call the tool if it actually needs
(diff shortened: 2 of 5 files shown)


allenhutchison:
@coderabbit review

allenhutchison:
Round 2 — addressed the one nitpick.

**Skip double `JSON.stringify` when nothing was truncated** (400c038) — `prepareHistory` was serializing the full history twice on every send to compute the delta, even on the common no-op case. Added the identity-check fast-path (`truncated !== history`); the helper already returns the input reference when there's nothing to truncate. Strengthened the relevant tests from `toEqual` to `toBe` so the optimization can't silently regress.

Calling this slightly more than "Low value" — for sessions with hundreds of KB of history, two stringify passes per send is real wasted CPU on the path where this entire feature is supposed to be a no-op.

allenhutchison:
Round 3 — addressed the outside-diff finding.

**Stale token estimate after truncation** (55a832b) — `estimatedTokens` was a snapshot of the previous turn's prompt size, used unmodified to gate compaction even after truncation shed bytes from this turn's outgoing history. On the turn truncation actually fires, that meant the threshold check could trigger compaction even though we'd be under threshold post-truncation.

Took a different path than the suggested `countTokens` API call: subtracted an estimated token reduction from the cached count using the standard 4-chars-per-token heuristic (the same one used elsewhere in this file for Ollama). Free, no roundtrip, accurate enough to skip the spurious compaction.

Test asserts that with a cached estimate just over 200K and a 600 KB tool-result payload that gets truncated, compaction is skipped and `countTokens` is never called.

PR #773, opened 2026-05-08
refactor(context): two-phase compaction (truncate first, summarize second)

allenhutchison:
## Summary

Follow-up to #770 (tool-result truncation). The shipped truncation pass ran on every send, even below threshold — and that costs users money. Modifying older history bytes invalidates Gemini's implicit prefix cache from the modified point forward; bytes that would have been served at the cached rate get billed at full rate for the turn truncation fires. Below threshold, that cache-miss tax is pure waste — the session didn't need any size relief.

This PR restructures `ContextManager.prepareHistory` as a two-phase pipeline that only churns the cache when we'd be paying that cost anyway:

1. **Below threshold** → history returned unchanged. Implicit cache stays valid; subsequent turns keep their cached-token discount.
2. **Over threshold, phase 1** → truncate old tool-result payloads (the existing helper). Re-estimate using the chars-per-token heuristic.
3. **If phase 1 brought us back under threshold** → done. Skip summarization entirely.
4. **Still over** → phase 2: existing summarization compaction, running on the truncated history.

The cache-miss point is the same whether we land in phase 1 or phase 2 — both modify older history. So when phase 1 fires we're paying a cost we'd otherwise pay for phase 2, and we get to skip the LLM summarization roundtrip plus keep recent turns verbatim instead of summarized.

## Changes

- `src/services/context-manager.ts` — restructured `prepareHistory`. Truncation now lives inside the over-threshold branch, with a re-evaluation gate that short-circuits when phase 1 alone is sufficient. Comments call out the cache-cost rationale explicitly so the next person to touch this code doesn't undo it.
- `test/services/context-manager.test.ts` — three new cases:
  - Phase 1 suffices when bloat is concentrated in one big tool result (asserts `wasCompacted=false`, oldest tool result truncated, `countTokens` never called).
  - Under threshold, even big old tool results are left intact (asserts input array reference returned unchanged + content fully preserved — pins the cache-preservation contract).
  - Phase 2 fires when phase 1 alone is insufficient (asserts `wasCompacted=true` on history dominated by text turns rather than tool-result bloat).
- `docs/reference/settings.md` — rewrote the subsection as "Two-phase compaction" with an explanation of the cache trade-off and why truncation is gated on the threshold.

## Verification

- `npm test` — 1671 tests passing (+2 in `context-manager.test.ts`).
- `npm run typecheck:test` — clean.
- `npm run build` — clean.

## Screenshots / Screencast

N/A — internal refactor.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#763)
- [x] I have tested this change on Desktop — pure-logic refactor with unit coverage; live integration impact will be measured via the eval harness when next run.
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — N/A, ContextManager is platform-neutral.
- [x] Documentation has been updated — `docs/reference/settings.md` rewritten "Two-phase compaction" section.

### AI-Generated Code

## Summary by CodeRabbit

* **Documentation**
  * Updated context management docs to describe the new two-phase compaction strategy for conversation history.

* **Improvements**
  * Two-phase compaction: first truncate oversized historical tool outputs, then run summarization only if needed.
  * Avoids unnecessary truncation when history is already below the compaction threshold to preserve cached behavior.

* **Tests**
  * Added/updated tests covering phase-1 truncation, threshold-preservation, and phase-2 summarization behavior.

### DIFF
--- a/docs/reference/settings.md
+++ b/docs/reference/settings.md
@@ -209,9 +209,17 @@ Context management automatically monitors and controls conversation size to prev
 - **How it works**: When conversation tokens exceed this percentage, older turns are summarized and replaced with a compact summary while preserving recent messages
 - **Hard ceiling**: Aggressive compaction triggers at 80% of the input limit to prevent API errors
 
-### Tool-result truncation in history
+### Two-phase compaction
 
-In addition to compaction, the plugin sheds bloat from older tool-result turns before each request. When a tool (typically `read_file`) returns a payload larger than ~4 KB, the response in conversation history is replaced with a small `{ truncated: true, truncatedFrom: N, note: "..." }` marker once it falls outside the most recent two tool-result turns. This keeps the agent's reasoning intact for the current and immediately previous tool call while preventing the long tail of past tool responses from dominating every subsequent prompt. Re-issuing the original tool call brings the full output back when needed. The behavior is always-on and not currently exposed as a setting.
+When a session crosses the compaction threshold the plugin runs a cheaper pass before reaching for full summarization:
+
+1. **Phase 1 — tool-result truncation.** Walks history and replaces oversized (>4 KB) `functionResponse` payloads in older turns with a small `{ truncated: true, truncatedFrom: N, note: "..." }` marker. The most recent two tool-result turns are kept intact so the agent reasoning across recent tool calls still has the full text. This is purely structural — no LLM call, no extra tokens spent.
+2. **Re-evaluation.** If phase 1 freed enough room to put us back under the threshold (e.g., a single 600 KB `read_file` was responsible for most of the bloat), the request goes out with the truncated history and phase 2 is skipped entirely.
+3. **Phase 2 — summarization.** Only fires when truncation alone wasn't enough. Older turns are summarized via an LLM call into a single context-summary entry preserving recent messages.
+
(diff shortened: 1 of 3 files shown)


allenhutchison:
Round 2 — addressed both findings.

**Outside-diff Major: "Run phase 1 before fast-path returns"** — disagreeing. The premise is that `prepareHistory` runs during tool-loop iterations and could see brand-new oversized history before `estimatedTokens` reflects it. That's not how the loop works: `prepareHistory` is called exactly once per user turn at \`agent-view-send.ts:306\`. The agent-loop's follow-up calls bypass it entirely (they go through \`buildToolHistoryTurns\` → direct API call). So at every entry to `prepareHistory`, `estimatedTokens` and `conversationHistory` reflect the same point-in-time (end of prior turn).

Implementing the suggestion would re-introduce the cache-miss-while-under-threshold cost this PR is specifically fixing. Below the threshold, modifying older history bytes invalidates Gemini's prefix cache for no benefit — we're not in size-pressure territory yet. The new test at `test/services/context-manager.test.ts` "does not truncate under threshold even when big tool results are present" pins exactly this contract.

(If a follow-up PR wires `prepareHistory` into mid-loop iterations, the picture changes — but that's a separate change with its own design considerations.)

**Nitpick: `wasCompacted` flag semantics** (0f76077) — clarified the JSDoc on `CompactionResult.wasCompacted` to state explicitly that it's true only when phase 2 summarization ran, paired with `summaryText`. The single consumer (\`agent-view-send.ts:309\`) already gates on the conjunction `wasCompacted && summaryText`, so phase 1 (no summary) correctly stays silent in user-facing notifications. Skipped adding a `wasTruncated` field — no caller currently needs it, so it'd be speculative API.

PR #784, opened 2026-05-10
fix(agent): preserve perTurnContext across follow-up requests

allenhutchison:
## Summary

Drag-and-drop / @-mention context files were rendered into `perTurnContext` on the initial model call but disappeared from every follow-up after a tool fired, because `buildFollowUpRequest` only plumbed `customPrompt` and tool-permission fields through.

Two consequences for the user:

1. Once the model called any tool, its system prompt was rebuilt without the rendered context-file content (and without `projectInstructions` / `projectSkills`), so the model couldn't reference the dragged file anymore — it tended to re-read the same file via tools, which looked like the chip was being ignored.
2. Gemini's implicit prefix cache keys on the exact system-prompt bytes. Stripping these fields between the initial call and the follow-up changed the prefix shape mid-turn, forcing a cache miss on every follow-up in long tool chains.

This was provider-agnostic — same code path on both Gemini and Ollama. It just became visible on Ollama because gemma readily falls back to `read_file` and on Gemini it mostly cost cache hits silently.

## Verification

Live capture against `gemma4:latest`:

| | Before | After |
|---|---|---|
| Initial system-prompt size | 31635 | 31635 |
| Follow-up system-prompt size | 25020 (-6.6 KB) | 31635 |
| `## Turn Context` heading in follow-up | absent | present |
| Rendered context-file contents in follow-up | absent | present |
| Initial vs. follow-up byte-identical | no | **yes** |

End-to-end eval (`context-from-shelf`) on `gemma4:latest`: SOLVED in 1 turn / 0 tool calls.

## Changes

**Fix**

- New `PerTurnContext` interface in `agent-view-tool-followup.ts` — the four fields that must stay byte-stable across the turn.
- `buildFollowUpRequest` and `buildRetryRequest` now extend `PerTurnContext` and pass the fields through.
- `AgentLoopOptions` gains `perTurn?: PerTurnContext`; `AgentLoop.run` spreads it onto the helper params.
- `AgentViewTools.handleToolCalls` accepts and forwards `perTurn` to the loop.
- `agent-view-send.ts` builds the `perTurn` object once and passes it to both `handleToolCalls` call sites (streaming + non-streaming).

`inlineAttachments` deliberately not threaded through — they belong to the user's original turn and would create empty-content user messages on follow-ups.

**Regression coverage** (added in second commit)

Three layers so this can't silently regress again:

1. `test/agent/agent-loop.test.ts` — 3 tests covering follow-up propagation, empty-response retry propagation, and multi-iteration tool chains. All four perTurn fields verified.
2. `test/api/ollama-client.test.ts` — asserts `perTurnContext` lands in the system message under `## Turn Context` on the wire to the Ollama daemon.
3. `test/api/gemini-client.test.ts` — captures the SDK's `generateContent` params and asserts `perTurnContext` is in `config.systemInstruction`. Also pins the negative case so the section doesn't appear when no context is set (cache-prefix stability).
4. `evals/tasks/context-from-shelf.json` (new) — full end-to-end task. New `addContextFiles()` driver helper threads a vault file onto the agent shelf via the same `addContextFileToShelf` API drag-drop uses, then asks a question answerable only from that file. Forbids `read_file` / `find_files_*` / `list_files` so a model that ignores the in-context content has nowhere to fall back.

Together: user message → agent-view-send → AgentLoop → {Gemini|Ollama}Client → follow-up → retry, all instrumented.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (interactively diagnosed)
- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards) — pure agent-loop plumbing, no platform-gated code touched
- [x] Documentation has been updated (if applicable) — internal plumbing only, no user-facing doc surface

### AI-Generated Code

## Test plan

- [x] `npm test` — 1685 passing (3 agent-loop + 1 Ollama + 2 Gemini new)
- [x] `npm run build` — clean
- [x] `npm run format-check` — clean
- [x] Live verification: dragged file → asked question → captured Ollama payload shows initial + follow-up system prompts byte-identical (31635 == 31635)
- [x] `npm run eval -- --task=context-from-shelf` — SOLVED in 1 turn / 0 tool calls on `gemma4:latest`

### DIFF
--- a/evals/fixtures/context-from-shelf/places-i-want-to-visit.md
+++ b/evals/fixtures/context-from-shelf/places-i-want-to-visit.md
@@ -1,0 +1,26 @@
+# Places I Want to Visit
+
+A bucket-list of destinations I'd love to see in my lifetime.
+
+## Confirmed Top Three
+
+- The Great Wall of China
+- Machu Picchu in Peru
+- Petra in Jordan
+
+## Other Dream Destinations
+
+- The Northern Lights from Tromsø, Norway
+- Angkor Wat in Cambodia
+- Gorilla trekking in Rwanda
+
+## Domestic Trips Planned for 2027
+
+- Hike a section of the Appalachian Trail
+- Visit Glacier National Park in Montana
+
+## Notes
+
+I want to do the Great Wall before I turn 50. Petra needs to be combined with
+a side trip to Wadi Rum. Machu Picchu requires permits booked at least four
+months in advance.
--- a/evals/lib/obsidian-driver.mjs
+++ b/evals/lib/obsidian-driver.mjs
@@ -126,6 +126,45 @@ export async function createSession(title) {
 }
 
 /**
+ * Populate the current agent session's shelf with the given vault paths so
+ * subsequent turns receive the rendered file content via `perTurnContext`,
+ * mirroring the user-facing drag-and-drop / @-mention flow.
+ *
+ * Must be called AFTER createSession() so `agentView.currentSession` and
+ * `agentView.shelf` reference the session under test. Throws if a path
+ * doesn't resolve to a vault TFile — that's an eval-task authoring error,
+ * not something the harness should silently paper over.
+ *
+ * Used by tasks that test context-chip behavior (e.g. "context-from-shelf"):
+ * the file lives on disk via setupFixtures, then this routine threads it
+ * into the agent context the same way the user would.
+ */
+export async function addContextFiles(paths) {
+	if (!Array.isArray(paths) || paths.length === 0) return;
+	const pathsLiteral = JSON.stringify(paths);
+	await obsidianEval(
+		`(async () => {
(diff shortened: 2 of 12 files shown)


allenhutchison:
Round 2 — addressed CodeRabbit feedback:

- **Move `PerTurnContext` out of the UI module** (`agent-loop.ts:9-14`): fixed in cdcd9f12e7cbbb0be48b32f67d4e2c0cdc1c214a. Promoted to `src/types/agent.ts` (existing home for agent-level types). `agent-loop.ts` now imports from there alongside `ChatSession`. The followup module re-exports the type so UI-internal imports keep working unchanged. The function helpers (`buildFollowUpRequest` etc.) still live in `agent-view-tool-followup.ts` — relocating those is a separate, larger refactor outside this fix's scope.

PR #786, opened 2026-05-10
feat(agent): record model name in session-history entries

allenhutchison:
## Summary

Session-history files now show which model produced each AI response. The Handlebars template at `src/history/templates/historyEntry.hbs` already renders a `| Model |` row when `entry.model` is set, and the round-trip parser at `src/agent/session-history.ts` already extracts it back — but every `role: 'model'` entry constructed by the agent view left the field undefined, so resumed sessions never displayed the model.

This is the same field that already shows Temperature and Top P when set; just plumbing the last piece through.

## Changes

Populate `model` at all five AI-entry construction sites:

- `src/ui/agent-view/agent-view-send.ts` — compaction summary, streaming-partial-before-tools, streaming-final, non-streaming-final
- `src/ui/agent-view/agent-view-tools.ts` — final response after tool loop

Each site uses the same expression that builds the request: `session.modelConfig?.model || settings.chatModelName`. Provider-agnostic — captures the requested model name, which is what both `GeminiClient` and `OllamaClient` receive on the request. User entries are unchanged (the template only renders the Model row when the field is set).

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — maintainer requested directly
- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards) — pure data-flow change in shared code, no platform-gated paths touched
- [x] Documentation has been updated (if applicable) — the per-message metadata block isn't documented anywhere; this populates a field the template already supported

### AI-Generated Code

## Test plan

- [x] `npm test` — 1685 passing
- [x] `npm run build` — clean (tsc + esbuild)
- [x] `npm run format-check` — clean
- [ ] Manual: send agent message, switch sessions, reopen — verify session file shows `| Model | <name> |` in the metadata block for assistant turns and is absent for user turns

### DIFF
--- a/src/ui/agent-view/agent-view-send.ts
+++ b/src/ui/agent-view/agent-view-send.ts
@@ -319,6 +319,7 @@ To reference an attachment in your response, use the path shown above.`;
 						message: `> [!info] Context Compacted\n> Older conversation turns have been summarized to maintain performance.\n\n${compactionResult.summaryText}`,
 						notePath: '',
 						created_at: new Date(),
+						model: modelName,
 					};
 					await this.ctx.displayMessage(compactionEntry);
 					await this.ctx.plugin.sessionHistory.addEntryToSession(currentSession, compactionEntry);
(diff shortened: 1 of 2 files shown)


allenhutchison:
@coderabbit review

PR #827, opened 2026-05-16
fix(evals): wait on turn events after dispatch

### LINKED ISSUE #778: evals: gemini-2.5-flash-lite reproducibly hangs the harness mid-sweep
## Motivation

Surfaced while landing the multi-model baselines (#774). `npm run eval -- --model=gemini-2.5-flash-lite` hangs partway through the 21-run sweep, both times in a different spot:

- First attempt: hung in `setupFixtures` for `multi-file-summary [run 1/3]` at the 9/21 mark.
- Second attempt: hung between `Turn completed.` and the verdict for `multi-hop-retrieval [run 3/3]` at the 14/21 mark.

In both cases the parent eval process sat alive at near-zero CPU; the CLI bridge from outside was responsive; manual equivalent calls returned correct values immediately. Killed both runs after 1h+ of no progress.

The other three models in the same session (`gemini-3-flash-preview`, `gemini-2.5-flash`, `gemini-2.5-pro`) all completed successfully (with one manual unstick on pro — see the broader CLI-bridge issue for that).

## Hypothesis

Likely tangled with the CLI-bridge hang issue (sibling issue), but reproducibility specific to lite suggests there's something model-specific too. Lite may produce a response shape (streaming-related? thinking-content related?) that interacts with `obsidianEval`'s child stdio differently than the larger models.

Worth checking whether lite's responses go through a different agent-loop code path — e.g., shorter response that triggers an empty-response retry, or a turn-end signal that the collector misses.

## Acceptance

- A full 21-run sweep against `gemini-2.5-flash-lite` finishes without manual intervention.
- Baseline can be `eval:bless`-ed and added to `evals/baselines/`.

## Code refs

- `src/agent/agent-loop.ts` — turn-end / empty-response paths
- `evals/lib/collector.mjs` — event capture
- `evals/lib/obsidian-driver.mjs` — `sendMessage`'s 300_000 ms outer timeout

ZardLi1115:
## Summary

Fixes #778 by removing the long-lived `obsidian eval` child from agent turns. The harness now dispatches the message with a short CLI call, then waits for collector `turnEnd` / `turnError` events under the task timeout budget.

## Changes

- Dispatch eval messages without awaiting the full model turn inside the Obsidian CLI process.
- Add a `turn-waiter` helper that polls collector events, tolerates transient collector read failures, and returns timeout state deterministically.
- Update eval docs to remove the `gemini-2.5-flash-lite` skip caveat and describe the current CLI-hang behavior.

### Required

- [ ] I have tested this change on Desktop (not run: this container does not have Obsidian/obsidian-cli available for a live eval sweep)
- [x] I have verified this change does not break Mobile (N/A: eval harness runs outside the Obsidian mobile app)

### AI-Generated Code

- [x] AI tool(s) used: Codex

## Tests

- `npm test -- test/evals/turn-waiter.test.ts test/evals/progress.test.ts test/evals/spawn-with-timeout.test.ts test/evals/scorer.test.ts`
- `npm run format-check`
- `npm run build`
- `npm test`
- `npm run lint` (passes with existing warnings unrelated to this PR)
- pre-push hook: `npm run build` and `npm test`

Not run: `npm run eval -- --model=gemini-2.5-flash-lite`; this environment has no `obsidian` CLI/live Obsidian instance configured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents long-lived CLI hangs by changing message dispatch to return immediately, enforcing shorter per-call timeouts, and treating stuck child processes as harness errors.

* **Documentation**
  * Clarified CLI timeout behavior and updated baseline-promotion guidance to mark CLI failures as ERROR and warn against blessing manually-killed baselines.

* **Tests**
  * Added unit tests for turn-completion polling and terminal-event detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/allenhutchison/obsidian-gemini/pull/827?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/evals/README.md
+++ b/evals/README.md
@@ -215,7 +215,7 @@ If you need to compare models, run them sequentially. A typical 21-run sweep tak
 
 ### CLI-bridge hangs (#776)
 
-`obsidian eval` child processes occasionally don't exit after their work completes — they sit in `S` state with 0 CPU. The next CLI call from the same harness queues behind them indefinitely. The harness's `execFile` `timeoutMs: 10_000` doesn't always fire.
+`obsidian eval` child processes have occasionally failed to exit after their work completed — they sat in `S` state with 0 CPU. The harness now avoids long-lived turn-driving CLI calls and uses a hard SIGTERM-to-SIGKILL timeout for every Obsidian CLI call, so these failures should settle as harness errors instead of wedging a sweep indefinitely.
 
 Symptoms in the log:
 
@@ -223,7 +223,7 @@ Symptoms in the log:
 - The harness's node process at near-zero CPU
 - A standalone `obsidian eval code="1+1"` from another shell **does** respond instantly (so the CLI itself is fine; the harness's specific child is wedged)
 
-Workaround until #776 is fixed:
+If a future run still appears stuck, check for stale children:
 
 ```bash
 # Find stuck children (zero CPU, alive for minutes)
@@ -233,7 +233,7 @@ ps aux | grep "obsidian eval" | grep -v grep
 kill -KILL <pid>
 ```
 
-The parent's `exec` promise will reject with `"Command failed"`, which `runTask`'s catch block records as **ERROR** and moves to the next run. **Do not bless** a baseline that includes a manual-kill ERROR — the verdict was caused by the harness, not the model. Rerun the whole sweep instead.
+The harness records CLI failures as **ERROR** and moves to the next run. **Do not bless** a baseline that includes a manual-kill ERROR — the verdict was caused by the harness, not the model. Rerun the whole sweep instead.
 
 ### Don't bless a corrupted run
 
@@ -265,7 +265,6 @@ If any of these show stale state, kill / clean before starting the new run. The 
 ### Model-specific caveats
 
 - **`*-latest` model pointers** (`gemini-flash-latest`, etc.) shift under us — Google may swap the underlying model on any given day. Don't bless a baseline against `-latest`; the comparison won't be stable. Use pinned IDs (`gemini-2.5-flash`, `gemini-2.5-pro`, etc.).
-- **`gemini-2.5-flash-lite`** reproducibly hangs the harness mid-sweep (#778). Skip it for now.
 
 ## Scoring
 
--- a/evals/lib/obsidian-driver.mjs
+++ b/evals/lib/obsidian-driver.mjs
@@ -190,7 +190,10 @@ export async function setupFixtures(files) {
 
 /**
  * Send a message to the agent via the programmatic API.
- * Returns immediately — use waitForTurnEnd() to wait.
+ * Returns after dispatch; the runner waits for `turnEnd` / `turnError` via
+ * the event collector. Do not await the full agent turn inside this eval call:
+ * keeping a CLI child open for the whole model request is what made sweeps
+ * vulnerable to mid-suite CLI bridge hangs (#778).
  */
 export async function sendMessage(text) {
 	// JSON.stringify produces a safely quoted/escaped JS string literal that can
@@ -199,10 +202,13 @@ export async function sendMessage(text) {
 	await obsidianEval(
 		`(async () => {
     const p = app.plugins.plugins['gemini-scribe'];
-    await p.agentView.sendMessageProgrammatically(${textLiteral});
+    const run = p.agentView.sendMessageProgrammatically(${textLiteral});
+    window.__evalInFlightSend = run.catch((err) => {
+      window.__evalLastSendError = err instanceof Error ? err.message : String(err);
+    });
     return '"sent"';
   })()`,
-		{ timeoutMs: 300_000 }
+		{ timeoutMs: 15_000 }
 	);
 }
 
--- a/evals/lib/obsidian-driver.mjs
+++ b/evals/lib/obsidian-driver.mjs
@@ -202,10 +202,14 @@ export async function sendMessage(text) {
 	await obsidianEval(
 		`(async () => {
     const p = app.plugins.plugins['gemini-scribe'];
+    const sendToken = (window.__evalLastDispatchToken ?? 0) + 1;
+    window.__evalLastDispatchToken = sendToken;
     window.__evalLastSendError = null;
     const run = p.agentView.sendMessageProgrammatically(${textLiteral});
-    window.__evalInFlightSend = run.catch((err) => {
-      window.__evalLastSendError = err instanceof Error ? err.message : String(err);
+    run.catch((err) => {
+      if (window.__evalLastDispatchToken === sendToken) {
+        window.__evalLastSendError = err instanceof Error ? err.message : String(err);
+      }
     });
     return '"sent"';
   })()`,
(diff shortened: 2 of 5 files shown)


allenhutchison (commented):
## Code review — decoupling dispatch from turn completion

Reviewed the full diff plus the surrounding harness and the agent event semantics this change relies on. **Solid, well-tested PR — the core approach is correct.** One should-fix item below; the rest are minor.

### Verified against the code

- `turnEnd` is emitted **once per user message**, from the `finally` block of `AgentViewSend.sendMessage()` (`src/ui/agent-view/agent-view-send.ts:582`) — *not* per agent-loop iteration. So `hasTerminalTurnEvent` matching the first `turnEnd` is correct; `AgentLoop` runs entirely inside the `try` and surfaces no extra terminal events.
- Any failure **inside** the main `try` (`agent-view-send.ts:198`) emits `turnError` (`:576`) then `turnEnd` (`:582`) — `waitForTurnCompletion` reliably terminates for that failure class.
- `obsidianEval` runs through `runWithTimeout` (`evals/lib/spawn-with-timeout.mjs`), which escalates SIGTERM→SIGKILL after a 1 s grace, so every `peekCollector` poll is hang-proof and the loop always makes progress. The README's "hard SIGTERM-to-SIGKILL timeout" claim is accurate.
- `readAndClearLastSendError`'s `JSON.parse(obsidianEval(... return JSON.stringify(x) ...))` round-trip matches the established `getSetting` pattern — correct.

### 1. Should fix — send-error check misses async pre-turn dispatch failures

`run.mjs` checks `readAndClearLastSendError()` **immediately after dispatch**, before `waitForTurnCompletion`. That catches a **synchronous** throw in `AgentViewSend.sendMessage()`'s sync prefix (lines 65–143): the `.catch()` microtask sets `window.__evalLastSendError` before the dispatch eval's CLI process exits.

It does **not** catch an **asynchronous** rejection that emits no terminal event. `sendMessage()` has a pre-`try` region (lines 64–197); `turnError`/`turnEnd` are only emitted from the `try` at line 198. An `await` that rejects in that pre-`try` region — most realistically `await this.ctx.displayMessage(userEntry)` at `agent-view-send.ts:196` — rejects `sendMessageProgrammatically` with **no `turnEnd`/`turnError`**. That rejection settles *after* the dispatch eval already returned `"sent"`, so the early `readAndClearLastSendError()` races it and usually reads `null`. `waitForTurnCompletion` then sees no terminal event, burns the full task budget (default 5 min), and the run is recorded **TIMEOUT** instead of **ERROR**.

On `master` the old `sendMessage` awaited the whole `sendMessageProgrammatically` promise, so it caught that rejection whenever it settled and reported ERROR. So this is a genuine (if narrow) regression — and it's the exact gap the earlier review comment described. That thread is currently marked resolved, but only the synchronous half is covered; consider reopening it.

Suggested fix — keep the early check (it's good, and faster than waiting the full budget for sync failures) and add a second read when the wait didn't complete:

```js
if (!waitResult.completed) {
    const lateSendError = await readAndClearLastSendError();
    if (lateSendError) {
        await cancelAgent();
        throw new Error(`sendMessageProgrammatically failed: ${lateSendError}`);
    }
    timedOut = true;
    // ...existing timeout handling
}
```

(Alternatively, fold an `__evalLastSendError` peek into `waitForTurnCompletion` and treat a non-null value as terminal.)

### 2. Minor — `window.__evalInFlightSend` is assigned but never read

In `obsidian-driver.mjs` `sendMessage`, the `.catch()` is what suppresses the unhandled rejection; storing its result on `window.__evalInFlightSend` serves no purpose. Drop the assignment, or add a comment if it's intentionally kept as a devtools handle.

### 3. Minor — `run.mjs` doesn't wire `onPollError`

`waitForTurnCompletion` exposes an `onPollError` hook, but `run.mjs` doesn't pass one. If `peekCollector` fails persistently, the loop spins silently for the whole budget (no progress lines, no warnings) before reporting TIMEOUT. Not a regression — the old poll loop also swallowed errors — but wiring a throttled `console.warn` would be a cheap observability win now that the hook exists.

### 4. Nit — `waitForTurnCompletion` doesn't guard `lastEvents`

`hasTerminalTurnEvent` defends `(events || [])`, but `lastEvents = events` does not. Harmless today (`peekCollector` always returns an array) — cosmetic only.

### 5. Doc — README drops the flash-lite caveat ahead of empirical proof

Removing the `gemini-2.5-flash-lite` "skip it" caveat is reasonable given the structural fix, but the PR notes no live sweep was run. Fine to merge — just flagging that the docs now assert behavior not yet observed end-to-end.

### Strengths

- Comments consistently explain *why*, not *what*.
- `waitForTurnCompletion` injects `now`/`sleep` for deterministic tests, and the `min(pollInterval, remaining)` clamp is covered (the `2_000` then `1_000` assertion).
- Clean removal of dead code (`TIMEOUT_SETTLE_MS`, the local `sleep`) with no orphaned references.

---

allenhutchison (commented):
## Follow-up review — changes since `507d424`

Re-reviewed the two new commits (`507d424` → `7387d58`). Both items from my previous review that needed a fix are now resolved.

### Resolved

**Should-fix — async dispatch failures misreported as TIMEOUT.** Fixed. `run.mjs` keeps the early `readAndClearLastSendError()` and adds a late re-check inside the `!waitResult.completed` branch before setting `timedOut`:

```js
if (!waitResult.completed) {
    const lateSendError = await readAndClearLastSendError();
    if (lateSendError) {
        await cancelAgent();
        throw new Error(`sendMessageProgrammatically failed: ${lateSendError}`);
    }
    timedOut = true;
    ...
}
```

Verified correct: a pre-`try` async rejection in `AgentViewSend.sendMessage()` emits no terminal event, so `waitForTurnCompletion` returns `completed: false`, and by then `window.__evalLastSendError` is reliably set → ERROR rather than TIMEOUT. Conversely `completed: true` implies a `turnEnd`/`turnError` fired (so no unhandled rejection) — gating the late check to the `!completed` path is sound.

**Minor — unused `window.__evalInFlightSend`.** Removed; the snippet is now a bare `run.catch(...)`.

### New code — dispatch-token guard

`sendMessage` now stamps each dispatch with an incrementing `window.__evalLastDispatchToken` and only records the error if the token still matches when the promise rejects — a stale rejection from a prior turn can't clobber the current turn's error slot. Logic is sound. Skipping CodeRabbit's suggested `Promise.resolve(...)` wrapper is the right call: `sendMessageProgrammatically` is `async`, so it already always returns a promise.

The still-open CodeRabbit thread on `evals/lib/obsidian-driver.mjs` (cross-turn misattribution) predates this commit — the current code already implements an equivalent token guard, so that thread looks stale.

### Still open — both optional, non-blocking

- **Minor** — `run.mjs` still doesn't pass `onPollError` to `waitForTurnCompletion`; a persistently failing `peekCollector` spins silently for the whole task budget. A throttled `console.warn` would close the observability gap.
- **Nit** — in `turn-waiter.mjs`, `lastEvents = events` lacks the `(events || [])` guard that `hasTerminalTurnEvent` has. Cosmetic — `peekCollector` always returns an array today.

### Verdict

The should-fix is properly closed and the new dispatch-token guard is correct. The two remaining items are a minor and a nit — neither blocking. Looks good to merge.

---

PR #835, opened 2026-05-16
fix(evals): allow judge key override

### LINKED ISSUE #788: evals: judge matchers fail on Ollama runs (no plugin apiKey → judge unavailable)
## What

Tasks that use `outputMatchers: [{ type: "judge", criteria: ... }]` always score `passed: true, solved: false` when the active provider is Ollama and the user has no Gemini API key configured in the plugin. The matcher evaluates as `pass: false` not because the model's answer is wrong, but because the judge function is unavailable.

Surfaced this on the gemma4-latest baseline (#787): `multi-file-summary` went 0/3 SOLVED across the 3-rep sweep, with all three runs flagged `solve_details.matchers_pass: false, judge_attempted: true, judge_available: false`.

## Where

- Judge construction: `evals/lib/judge.mjs:101` — `createJudge` returns `null` when `readApiKey()` returns null.
- Plugin apiKey getter: returns null on Ollama-only setups (no Gemini key needed for daily use).
- Run.mjs warns once at startup ("⚠ Tasks reference \`judge\` matchers but no Gemini API key is reachable; those matchers will fail.") but the warning is easy to miss in the eval setup output, and there's no per-task indication in the result file beyond the boolean flags above.

## Reproduction

Plugin set to `provider: ollama` with no Gemini API key configured:

```bash
npm run eval -- --task=multi-file-summary --repeat=1
# → PASSED (not solved): 3 turns, expected_tools_met=true,
#   forbidden_tools_clean=true, matchers_pass=false,
#   judge_attempted=true, judge_available=false
```

The model's response is fine — it summarizes the three fixture topics. The judge just never runs.

## Why this matters

- **Misleading baseline numbers.** The gemma4-latest baseline shows solve^3 = 62.5%, but the 0/3 on multi-file-summary is a harness-config artifact, not a model regression. A reader comparing this to a Gemini baseline (where the judge runs) gets a wrong picture of relative model quality.
- **No clean way around it today.** A user running Ollama-only has no reason to have a Gemini key configured in the plugin, but they DO have a reason to want judge-matcher tasks scored.
- **Can hide real regressions later.** If we ship a model change that breaks something the judge would catch, an Ollama-only sweep would still pass^3 = 100% and the regression slips by.

## Suggested direction (not prescriptive)

`createJudge({ apiKey })` already accepts an explicit override. Two reasonable fixes:

1. **Wire `EVAL_JUDGE_API_KEY` env var through `run.mjs`.** Pass it to `createJudge` ahead of the plugin-apiKey lookup. Document in `evals/README.md` and the eval-harness skill: "If you run Ollama-only sweeps, set EVAL_JUDGE_API_KEY=... so judge matchers still score."
2. **Make the judge-unavailable case louder.** Print a per-task `[judge unavailable]` annotation next to the `PASSED (not solved)` line and add a `judge_skipped` field to `solve_details` so eval-compare can warn instead of treating it as a real solve regression.

(1) is the actual fix; (2) prevents repeating today's confusion even after the fix lands.

## Out of scope

The `multi-file-summary` task itself is fine — gemma4 produces a coherent three-topic summary every run. No need to weaken or rewrite the rubric.

ZardLi1115:
## Summary

Fixes #788.

Allows Ollama-only eval runs to provide a Gemini judge key through `EVAL_JUDGE_API_KEY`, so judge matchers can run even when the plugin itself has no Gemini API key configured.

## Changes

- Pass `EVAL_JUDGE_API_KEY` from `evals/run.mjs` into `createJudge` before falling back to the plugin API key.
- Add `judge_skipped` reporting and `[judge unavailable]` verdict annotation when judge matchers cannot run.
- Document Ollama-only judge setup in `evals/README.md` and the eval-harness skill.
- Add eval matcher/scorer regression coverage for skipped judge reporting.

### Required

- [ ] I have tested this change on Desktop (not run in live Obsidian; eval harness requires a configured desktop vault and credentials, covered here by unit tests)
- [ ] I have verified this change does not break Mobile (N/A: eval harness is a Node/Obsidian desktop operator tool)

### AI-Generated Code

## Tests

- `npm test -- test/evals/judge.test.ts test/evals/matchers.test.ts test/evals/scorer.test.ts`
- `npm run format-check`
- `npm run build`
- `npm test`

## Summary by CodeRabbit

* **New Features**
  * Support for EVAL_JUDGE_API_KEY to supply judge credentials independently.

* **Improvements**
  * Eval results now include judge_skipped and related judge status fields; runs emit a "judge unavailable" verdict tag.
  * Ollama-only runs treat missing judge creds as a harness setup failure (not a model-quality regression).
  * Key-resolution order clarified (EVAL_JUDGE_API_KEY preferred).

* **Tests**
  * Updated tests to assert judge_skipped and related behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/allenhutchison/obsidian-gemini/pull/835?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


allenhutchison (changes requested):
Thanks for the fix — the `EVAL_JUDGE_API_KEY` wiring is correct, the `judge_skipped` flag is consistent across `matchers.mjs` / `scorer.mjs` / `run.mjs`, and adding the judge fields to the error-path `solve_details` is a nice catch that harmonizes the result schema. Tests and docs look good.

Requesting two changes before merge:

### 1. Extract a `taskHasJudgeMatcher(task)` helper (DRY)

`(task.outputMatchers || []).some((m) => m?.type === 'judge')` now appears 5 times across `run.mjs` and `scorer.mjs` — this PR adds 3 of them. In both the `scorer.mjs` fallback object and the `run.mjs` catch block, the `.some()` is evaluated **twice within the same object literal** (`judge_attempted` and `judge_skipped`).

Please extract a one-line helper, exported from `evals/lib/matchers.mjs`:

```js
export const taskHasJudgeMatcher = (task) => (task.outputMatchers || []).some((m) => m?.type === 'judge');
```

and derive `judge_skipped` from the already-computed `judge_attempted && !judge_available` rather than recomputing `.some()`.

### 2. Mention `EVAL_JUDGE_API_KEY` in the runtime "judge unavailable" warning

When `createJudge()` returns `null`, `run.mjs` warns:

> ⚠ Tasks reference `judge` matchers but no Gemini API key is reachable; those matchers will fail.

That message fires exactly when the operator would benefit from knowing about the new env var, but `EVAL_JUDGE_API_KEY` currently only appears in the docs. Please update the warning to point at it — e.g. append `(set EVAL_JUDGE_API_KEY or configure a plugin Gemini key)`.

---

**Non-blocking note:** an _invalid_ `EVAL_JUDGE_API_KEY` (or a transient network error during a judge call) still surfaces as a plain task failure rather than `judge_skipped` — `evaluateMatchers`'s catch sets `pass = false` with no "judge errored" signal. Out of scope for this PR, but a reasonable follow-up given the goal of separating setup failures from model regressions.

---

allenhutchison:
Thanks — both requested changes look good in `0a05824`:

1. `taskHasJudgeMatcher(task)` is extracted into `evals/lib/matchers.mjs` and reused at every former call site; the duplicated `.some()` expression (including the double-evaluation inside the `scorer.mjs`/`run.mjs` object literals) is gone, with `judge_skipped` derived from `judgeAttempted && !judgeAvailable`.
2. The "judge unavailable" warning now points operators at `EVAL_JUDGE_API_KEY`.

`test` and `lint` are green on the new head.

---

PR #840, opened 2026-05-16
feat: stabilize system prompt prefix caching (#763)

### LINKED ISSUE #763: Repeated Agents.md System prompt injection bloats every call

allenhutchison:
## Summary

Apply the established `turnPreamble` design pattern to stabilize Gemini's prompt prefix caching across agent turns, resolving the quadratic prompt parsing costs encountered in long chat or coding sessions.

Fixes #763

## Changes

- **Prompt Caching Optimization**: Refactored dynamic `perTurnContext` (carrying context-chip files and attachments metadata) out of `systemInstruction` and appended it directly to the user's conversation message parts.
- **Clean Prompt Signature**: Removed the `perTurnContext` block from `prompts/systemPrompt.hbs` and its rendering inside `gemini-prompts.ts`'s `getSystemPromptWithCustom`.
- **Static `systemInstruction` Guarantee**: Updated `GeminiClient` and `OllamaClient` to construct completely static system prompts throughout the session, guaranteeing a **100% prefix cache hit rate on the system instruction** across turns (~75% off input token cost!).
- **Correctness & Tool Follow-up Continuity**: Upgraded `buildToolHistoryTurns` in `agent-loop-helpers.ts` to accept optional `perTurnContext` and merge it alongside the user message text inside the spliced `user` history turn. This ensures that files context is preserved in history and remains fully visible to the model across subsequent tool calling iterations.
- **Test Alignment & Coverage**:
  - Updated `test/agent/agent-loop-helpers.test.ts` with comprehensive tests verifying combined user parts splicing.
  - Updated `gemini-prompts.test.ts` to match the getSystemPromptWithCustom signature.
  - Refactored provider tests in `gemini/client.test.ts` and `ollama/client.test.ts` to assert turn-specific file context resides inside the user message content instead of the system prompt.

### Required

- [ ] I have verified this change does not break Mobile (not applicable - purely internal provider and prompt wiring changes)
- [ ] Documentation has been updated (not applicable - purely internal caching refactoring and prompt wiring optimizations with no user-facing settings or behavior changes)

### AI-Generated Code

## Summary by CodeRabbit

* **Refactor**
  * Per-turn context is no longer embedded in system prompts; it's sent as user message content to AI providers to ensure consistent handling across requests.
  * Agent loop and tool-history logic updated to include per-turn context exactly once per turn, preventing duplication across follow-up iterations.

* **Tests**
  * Added and updated tests to verify single-instance context embedding and that follow-up/retry requests omit duplicated context.

[Review Change Stack](https://app.coderabbit.ai/change-stack/allenhutchison/obsidian-gemini/pull/840?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/prompts/systemPrompt.hbs
+++ b/prompts/systemPrompt.hbs
@@ -60,9 +60,3 @@ or a web search.
 
 {{{additionalInstructions}}}
 {{/if}}
-{{#if perTurnContext}}
-
-## Turn Context
-
-{{{perTurnContext}}}
-{{/if}}
--- a/src/agent/agent-loop-helpers.ts
+++ b/src/agent/agent-loop-helpers.ts
@@ -158,10 +158,19 @@ export function buildFunctionResponseParts(toolResults: ToolCallResultPair[]): P
 export function buildToolHistoryTurns(args: {
 	conversationHistory: Content[];
 	userMessage: string;
+	perTurnContext?: string;
 	toolCalls: ToolCall[];
 	toolResults: ToolCallResultPair[];
 }): Content[] {
-	const { conversationHistory, userMessage, toolCalls, toolResults } = args;
+	const { conversationHistory, userMessage, perTurnContext, toolCalls, toolResults } = args;
+
+	const userParts: Part[] = [];
+	if (userMessage && userMessage.trim()) {
+		userParts.push({ text: userMessage });
+	}
+	if (perTurnContext && perTurnContext.trim()) {
+		userParts.push({ text: perTurnContext });
+	}
 
 	const updated: Content[] = [
 		...conversationHistory,
@@ -169,10 +178,10 @@ export function buildToolHistoryTurns(args: {
 		{ role: 'user', parts: buildFunctionResponseParts(toolResults) },
 	];
 
-	if (userMessage && userMessage.trim()) {
+	if (userParts.length > 0) {
 		updated.splice(conversationHistory.length, 0, {
 			role: 'user',
-			parts: [{ text: userMessage }],
+			parts: userParts,
 		});
 	}
 
--- a/src/agent/agent-loop.ts
+++ b/src/agent/agent-loop.ts
@@ -179,6 +179,11 @@ export class AgentLoop {
 		let currentToolCalls = initialResponse.toolCalls ?? [];
 		let conversationHistory = initialHistory;
 		let userMessage = initialUserMessage;
+		// `perTurnContext` is a first-iteration input, like `userMessage`:
+		// `buildToolHistoryTurns` splices it into the user turn once, after which
+		// it lives in `conversationHistory`. Re-passing it on later iterations
+		// would splice the (potentially large) context payload in again each time.
+		let perTurnContext = perTurn?.perTurnContext;
 		let iterations = 0;
 		// Turn-scoped count of tool-loop-detector fires. Incremented per blocked
 		// call (each `ToolResult` with `loopDetected: true`). Once it reaches
@@ -236,7 +241,7 @@ export class AgentLoop {
 				const updatedHistory = buildToolHistoryTurns({
 					conversationHistory,
 					userMessage,
-					perTurnContext: perTurn?.perTurnContext,
+					perTurnContext,
 					toolCalls: currentToolCalls,
 					toolResults,
 				});
@@ -262,7 +267,7 @@ export class AgentLoop {
 			const updatedHistory = buildToolHistoryTurns({
 				conversationHistory,
 				userMessage,
-				perTurnContext: perTurn?.perTurnContext,
+				perTurnContext,
 				toolCalls: currentToolCalls,
 				toolResults,
 			});
@@ -301,7 +306,10 @@ export class AgentLoop {
 				}
 				currentToolCalls = followUpResponse.toolCalls;
 				conversationHistory = updatedHistory;
+				// Both are now embedded in `updatedHistory`; clearing them stops the
+				// next iteration from splicing a duplicate user/context turn.
 				userMessage = ''; // Empty on follow-up — tool results are already in history
+				perTurnContext = undefined;
 				continue;
 			}
 
--- a/src/agent/agent-loop.ts
+++ b/src/agent/agent-loop.ts
@@ -236,6 +236,7 @@ export class AgentLoop {
 				const updatedHistory = buildToolHistoryTurns({
 					conversationHistory,
 					userMessage,
+					perTurnContext: perTurn?.perTurnContext,
 					toolCalls: currentToolCalls,
 					toolResults,
 				});
@@ -261,6 +262,7 @@ export class AgentLoop {
 			const updatedHistory = buildToolHistoryTurns({
 				conversationHistory,
 				userMessage,
+				perTurnContext: perTurn?.perTurnContext,
 				toolCalls: currentToolCalls,
 				toolResults,
 			});
--- a/src/api/providers/gemini/client.ts
+++ b/src/api/providers/gemini/client.ts
@@ -236,7 +236,6 @@ export class GeminiClient implements ModelApi {
 				agentsMemory,
 				availableSkills,
 				extReq.projectInstructions,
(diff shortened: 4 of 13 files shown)


allenhutchison:
## Follow-up fix: per-turn context was being duplicated (commit ed6b40d)

QA-testing this branch surfaced a bug in the tool-execution path, now fixed in `ed6b40d`.

### The bug

`buildToolHistoryTurns` (added by this PR) splices `perTurnContext` into the user turn of the history it builds. But `buildFollowUpRequest` / `buildRetryRequest` *also* set `perTurnContext` on the request, so `GeminiClient.buildContents` appended a **second copy**. Captured wire payload of a tool follow-up:

```
user  [ userMessage , perTurnContext ]      ← spliced into history
model [ functionCall: list_files ]
user  [ functionResponse: list_files ]
user  [ perTurnContext ]                    ← appended AGAIN by buildContents
```

A second facet: `AgentLoop` passed `perTurnContext` to `buildToolHistoryTurns` on *every* iteration, so a multi-tool loop re-spliced it once per iteration. Both send the full context payload extra times on every tool turn — undercutting the prefix-caching win this PR exists for, in exactly the tool-heavy coding sessions #763 is about.

### The fix

- `buildFollowUpRequest` / `buildRetryRequest` no longer set `perTurnContext` on the request — it is already embedded in `updatedHistory` via `buildToolHistoryTurns`. Session-static fields (`projectInstructions`, `projectSkills`, `sessionStartedAt`) still thread through.
- `AgentLoop` treats `perTurnContext` as a first-iteration input like `userMessage`: passed to `buildToolHistoryTurns` once, then cleared so later iterations don't re-splice.

### Heads-up on the test changes

The `per-turn context propagation` block in `test/agent/agent-loop.test.ts` previously **asserted the duplicated behavior** (`followUpRequest.perTurnContext` set). Those three tests are rewritten to assert the corrected behavior — context absent from the request, present exactly once in `conversationHistory`, including across a multi-iteration loop. New `test/ui/agent-view/agent-view-tool-followup.test.ts` locks the request-builder contract. Worth a careful look since the assertions flipped.

### Verification

Full suite passes (2835). Verified live against a real Obsidian instance with a `fetch` interceptor: the static `systemInstruction` is byte-identical across turns (71% cache hit on turn 2), and the tool follow-up now carries `perTurnContext` exactly once.

Not live-tested: image/binary attachments and the Ollama provider (both structurally unchanged by the fix; covered by existing unit tests).

PR #877, opened 2026-05-22
refactor: extract JsonSidecarStateStore + definition-file helpers

### LINKED ISSUE #864: dry-violation: hook-manager and scheduled-task-manager duplicate JSON-sidecar state + file-parsing scaffolding
## What

`src/services/hook-manager.ts` (1008 lines) and `src/services/scheduled-task-manager.ts` (940 lines) are sibling service-layer managers built on the same pattern: each discovers markdown definition files under a `[state-folder]/<Feature>/` directory, parses YAML frontmatter into a typed definition, serializes definitions back to markdown, and persists volatile per-entry runtime state to a JSON sidecar file. They were written largely in parallel and have accumulated substantial structural duplication.

This is the service-layer analogue of #815, which found the same parallel-design duplication between `hook-management-modal.ts` and `scheduler-management-modal.ts` and resolved it with a shared `ManagementModalBase`. The *manager* layer underneath those modals has no equivalent shared base, so the same fix that was applied to the UI is still missing one layer down.

## Evidence

Line numbers as of master:

**`loadState` / `saveState` — byte-identical apart from the state type cast, the log prefix, and the caught-error variable name:**

- `hook-manager.ts:986–1008` vs `scheduled-task-manager.ts:918–940`

**`parseHookFile` / `parseTaskFile` — same skeleton:** `metadataCache.getFileCache(file)?.frontmatter` guard → `vault.read(file)` + `findFrontmatterEndOffset` + body-slice → `parseToolPolicyFrontmatter(...) ?? migrateLegacyToolCategoryArray(...)` → build the typed object → **near-identical legacy `enabledTools` auto-migration block** (`if (frontmatter.enabledTools !== undefined && frontmatter.toolPolicy === undefined) { try { vault.modify(...) } catch { logger.warn(...) } }`):

- `hook-manager.ts:882–943` (migration block `932–940`) vs `scheduled-task-manager.ts:833–874` (migration block `863–871`)

**Folder / state-path getter trio:**

- `hook-manager.ts:281` `hooksFolder`, `:285` `runsFolder`, `:289` `stateFilePath`
- `scheduled-task-manager.ts:278` `scheduledTasksFolder`, `:282` `runsFolder`, `:286` `stateFilePath`

**Orphan-state purge** (drop sidecar state entries whose definition file no longer exists) exists in both: `scheduled-task-manager.ts:819–826`, and the equivalent `delete this.state[slug]` cleanup in `hook-manager.ts:876`.

## Proposed unit of work

- Introduce a generic state-persistence helper — e.g. `JsonSidecarStateStore<TState>` under `src/services/` — that owns `loadState`/`saveState` (`exists` → `read` → `JSON.parse` with graceful fallback; `write` with error logging), parameterized by the sidecar path and a log prefix. Both managers hold one instance instead of hand-rolling the methods.
- Pull the shared definition-file parsing skeleton into a small helper (the frontmatter-guard → body-extraction → tool-policy resolution → legacy-`enabledTools` auto-migration sequence). The feature-specific frontmatter-field mapping stays in each manager's `parseHookFile` / `parseTaskFile`.
- Consider a thin `FeatureDefinitionManager` base (or composition) for the `runsFolder` / `stateFilePath` getters and the orphan-state purge — mirroring how #815's `ManagementModalBase` consolidated the modal scaffolding.
- Pure code-motion refactor: no behavior change. `npm test` must stay green; extend the existing manager test suites if present.

## Out of scope

- Splitting either file to get under the 700-line threshold — tracked separately in #801. (This refactor will shrink both files as a side effect, but file-size reduction is not the goal here.)
- The modal-layer duplication — already resolved in #815.
- Changing the on-disk format of definition files or sidecar state.
- Behavior changes to debounce/cooldown/loop-detection (hooks) or schedule parsing (tasks).

---

allenhutchison:
## Summary

`hook-manager.ts` and `scheduled-task-manager.ts` were written in parallel and hand-rolled the same JSON-sidecar state I/O and definition-file parsing scaffolding. This pulls the duplicated pieces into a shared `src/services/feature-definition.ts` module. Pure code-motion — no behavior change.

Closes #864

## Changes

- New `src/services/feature-definition.ts` with shared helpers:
  - `JsonSidecarStateStore<T>` — load/save of the per-manager state JSON sidecar (the previously byte-identical `loadState`/`saveState`)
  - `extractMarkdownBody` — frontmatter-aware body extraction
  - `resolveFeatureToolPolicy` — `toolPolicy:` block with legacy `enabledTools:` fallback
  - `migrateLegacyEnabledTools` — in-place legacy-frontmatter rewrite
  - `purgeOrphanState` — drop state entries with no matching definition file
- `hook-manager.ts` and `scheduled-task-manager.ts` now delegate to those helpers (both files shrank ~14 lines; net −28 across the two).
- New `test/services/feature-definition.test.ts` — 16 cases covering the extracted module.

### Notes / scope

- `migrateLegacyEnabledTools` returns `undefined` **synchronously** for the no-op case, so a parse with nothing to migrate adds no extra `await` tick. This keeps `parseTaskFile`'s microtask timing identical (the scheduled-task hot-reload tests count microtask flushes).
- No shared `FeatureDefinitionManager` base class was introduced — composition via the helpers above covers the duplication without an inheritance layer. The 3 folder/state-path getters were left as-is: they are one-liners over already-named constants, and renaming the public `hooksFolder` / `scheduledTasksFolder` getters wasn't worth the churn.
- No user-facing, settings, or command/API changes, so no documentation updates were required.

## Verification

- `npm test` — 2861 passed (123 files), incl. the new `feature-definition` suite and the existing `hook-manager` / `scheduled-task-manager` suites unchanged.
- `npm run build` — `tsc -noEmit` clean, esbuild bundle clean.
- `npm run lint` — 0 errors (7 pre-existing warnings, all in unrelated files).

### Required

- [x] I have verified this change does not break Mobile (pure code-motion; no platform-specific logic touched)
- [x] Documentation has been updated (if applicable) — N/A, internal-only refactor

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_015Afp4KAEm8th93D33TrZPh)_

## Summary by CodeRabbit

* **Refactor**
  * Consolidated feature-definition and runtime state utilities across the app for more consistent handling of feature settings and tool policies.
  * Managers now persist runtime state via a dedicated sidecar mechanism, automatically purge stale entries, and centrally handle legacy tool-policy migrations and markdown prompt extraction.
* **Tests**
  * Added comprehensive tests covering state persistence, markdown extraction, policy resolution, legacy migration, and orphan-state purging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/allenhutchison/obsidian-gemini/pull/877?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/services/feature-definition.ts
+++ b/src/services/feature-definition.ts
@@ -1,0 +1,136 @@
+import type { TFile } from 'obsidian';
+import type ObsidianGemini from '../main';
+import { findFrontmatterEndOffset } from './skill-manager';
+import { FeatureToolPolicy, parseToolPolicyFrontmatter } from '../types/tool-policy';
+import { migrateLegacyToolCategoryArray } from './feature-policy-yaml';
+
+/**
+ * Shared scaffolding for the markdown-defined feature managers — HookManager
+ * and ScheduledTaskManager. Both discover `<slug>.md` definition files under a
+ * folder in the plugin state directory, parse YAML frontmatter into a typed
+ * definition, and persist volatile per-entry runtime state to a JSON sidecar.
+ * The helpers below are the parts of that pattern that are identical between
+ * the two managers; the feature-specific frontmatter-field mapping stays in
+ * each manager's own `parseHookFile` / `parseTaskFile`.
+ */
+
+/**
+ * Reads and writes a JSON sidecar state file (a `Record<slug, EntryState>`).
+ * Owns only the I/O: `load` tolerates a missing or corrupt file by returning an
+ * empty map, and `save` logs (but never throws) on write failure so a transient
+ * disk error cannot break a hook fire or a scheduler tick.
+ *
+ * The owning manager keeps the live state object — this store is purely the
+ * persistence layer. The sidecar path is resolved lazily on every call because
+ * it depends on `settings.historyFolder`, which can change at runtime.
+ */
+export class JsonSidecarStateStore<T extends object> {
+	/**
+	 * @param plugin       Plugin instance — provides the vault adapter and logger.
+	 * @param resolvePath  Returns the current sidecar file path. Invoked on every
+	 *                     load/save so a `historyFolder` change is picked up.
+	 * @param logPrefix    Bracketed manager tag for log lines, e.g. `[HookManager]`.
+	 */
+	constructor(
+		private readonly plugin: ObsidianGemini,
+		private readonly resolvePath: () => string,
+		private readonly logPrefix: string
+	) {}
+
+	/** Load the sidecar state, falling back to an empty map when absent or corrupt. */
+	async load(): Promise<T> {
+		const path = this.resolvePath();
+		try {
+			const exists = await this.plugin.app.vault.adapter.exists(path);
+			if (!exists) return {} as T;
+			const raw = await this.plugin.app.vault.adapter.read(path);
+			return JSON.parse(raw) as T;
+		} catch (err) {
+			this.plugin.logger.warn(`${this.logPrefix} Failed to load state, starting fresh:`, err);
+			return {} as T;
+		}
+	}
+
+	/** Persist the sidecar state. Logs but does not throw on write failure. */
+	async save(state: T): Promise<void> {
+		const path = this.resolvePath();
+		try {
+			await this.plugin.app.vault.adapter.write(path, JSON.stringify(state, null, 2));
+		} catch (err) {
+			this.plugin.logger.error(`${this.logPrefix} Failed to save state:`, err);
+		}
+	}
+}
+
+/**
+ * Extract the markdown body — everything after the YAML frontmatter block —
+ * trimmed. Returns the whole trimmed content when there is no frontmatter.
+ */
+export function extractMarkdownBody(content: string): string {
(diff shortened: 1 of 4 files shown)


allenhutchison:
Thanks for the review. Status on the two findings:

- **Validation (line 50)** — fixed in 3619e54. `JsonSidecarStateStore.load()` now rejects non-object JSON (null / array / primitive) and falls back to the empty-map / warn path. Added a regression test covering all three cases.
- **`processFrontMatter` nitpick (line 98–118)** — skipping in this PR. Switching the legacy migration from `vault.modify(serialize())` to `processFrontMatter` would change the on-disk YAML format (different key ordering/quoting, and we'd lose the "omit defaults" behavior of the current `serializeHook` / `serializeTask`). Issue #864 explicitly lists *"Changing the on-disk format of definition files"* as out of scope, and the original migration code already used `vault.modify`, so this is a pre-existing pattern rather than a regression. Happy to file a follow-up issue if you'd like to pursue the broader change separately.

---

PR #1005, opened 2026-06-21
refactor(services): extract shared FailurePauseTracker (#994)

### LINKED ISSUE #994: dry-violation: HookManager and ScheduledTaskManager carry parallel failure-pause state machines
## What

`HookManager` (1016 lines) and `ScheduledTaskManager` (951 lines) both implement the same "auto-pause after N consecutive failures" state machine over their per-entry sidecar state, with the same constant (`MAX_CONSECUTIVE_FAILURES = 3`) declared twice. They also have parallel `recordFailure`-style branches inside their run paths, parallel `recordSuccess` resets, and parallel `resetHook`/`resetTask` user-facing "clear the error" methods.

Issue #864 already extracted JSON-sidecar state + file-parsing scaffolding into shared utilities. This is the next layer of duplication on top of that base: the **error-recovery behavior** that uses those sidecars. Both managers walk the same prev-state → bump counter → derive `pausedDueToErrors` → save → log ladder, and both expose the same "reset failures" semantics with one entity-specific patch (a fresh `nextRunAt` for tasks).

The smell: a fix to the pause/log policy (e.g. tightening "3 consecutive" to "3 within 24h", or surfacing a `Notice` when something pauses) has to be applied in two places and the constants kept in lockstep manually. The next system that grows a failure-pause behavior (custom prompts? MCP servers?) will copy this a third time.

## Evidence

- `src/services/hook-manager.ts:37` — `const MAX_CONSECUTIVE_FAILURES = 3;` (module-scope).
- `src/services/scheduled-task-manager.ts:26` — `const MAX_CONSECUTIVE_FAILURES = 3;` (same value, redeclared).
- `src/services/hook-manager.ts:782–800` — `recordFailure(slug, error)` — bump counter, compute `pausedDueToErrors`, splat-merge new state, save, branch warn-vs-error log.
- `src/services/scheduled-task-manager.ts:789–807` — same ladder inline inside the per-task `runOnce` catch block (same constant, same log shape modulo `[ScheduledTaskManager]` vs `[HookManager]` prefix and the message string).
- `src/services/hook-manager.ts:771–780` — `recordSuccess(slug)` — clear `lastError`, `consecutiveFailures: 0`, `pausedDueToErrors: false`.
- `src/services/scheduled-task-manager.ts:776–786` — same clear-on-success, inline inside `runOnce`, plus an entity-specific `lastRunAt: new Date().toISOString()` patch.
- `src/services/hook-manager.ts:377–387` — `resetHook(slug)` — user-facing "clear failure state."
- `src/services/scheduled-task-manager.ts:631–641` — `resetTask(slug)` — same shape, plus the entity-specific `nextRunAt: new Date().toISOString()` patch.

Both managers also have `pausedDueToErrors` early-exit checks that read identically (lines 599–605 / 463–469 — the "is paused, skipping" log + return).

## Proposed unit of work

Introduce a small `FailurePauseTracker<TState extends FailurePauseState>` helper (e.g. `src/services/failure-pause-tracker.ts`) that owns the ladder:

- `FailurePauseState` is the shared shape: `{ lastError?: string; consecutiveFailures?: number; pausedDueToErrors?: boolean; }`.
- The tracker takes `(getState, setState, logger, label)` plus the max-failures constant (defaulting to a shared `MAX_CONSECUTIVE_FAILURES` exported from this module so callers can import the same `3` instead of redeclaring it).
- Methods: `recordSuccess(slug, patch?)`, `recordFailure(slug, error, patch?)`, `reset(slug, patch?)`, `isPaused(slug)`. The optional `patch` is how callers inject entity-specific fields (`lastRunAt`, `nextRunAt`, etc.) without the tracker knowing about them.
- Pure delegation — the tracker doesn't own persistence; callers wire `setState` to whatever `saveState` they already have. This keeps both managers' existing sidecar plumbing untouched.

Then have both managers delegate:

- `HookManager.recordFailure` / `recordSuccess` / `resetHook` and the inline `runOnce` blocks in `ScheduledTaskManager` all become tracker calls. The two logger messages keep their `[HookManager]` / `[ScheduledTaskManager]` prefixes via the `label` parameter.
- Delete both module-scope `MAX_CONSECUTIVE_FAILURES` declarations in favor of the shared export.

Tests:

- `test/services/failure-pause-tracker.test.ts` covering the ladder transitions (first-failure → counter=1, max-failures → `pausedDueToErrors`, success → reset, reset-with-patch behavior).
- Keep the existing `test/services/hook-manager.test.ts` and `test/services/scheduled-task-manager.test.ts` assertions about end-to-end pause behavior — they should still pass against the delegating implementations.

## Out of scope

- Changing the pause policy itself (still "3 consecutive failures" — if a maintainer wants a different policy that's a separate issue).
- Touching the early-exit "is paused, skipping" log sites (lines 599–605 / 463–469). They read the state directly via `isPaused` once the tracker exists, but the log message is entity-specific and best left in place; this issue is about the write side.
- Generalizing to a third subsystem (MCP servers, custom prompts, etc.) — extract for the two existing call sites, don't speculate.
- Moving the persistent state shape itself. Each manager keeps its own state file and sidecar format; the tracker is over the in-memory record only.

---

saheersk:
## Summary

`HookManager` and `ScheduledTaskManager` both implemented the same
"auto-pause after N consecutive failures" state machine over their
per-entry JSON sidecar state, with `MAX_CONSECUTIVE_FAILURES = 3`
declared twice and parallel `recordFailure` / `recordSuccess` /
`resetHook` vs `resetTask` logic. This extracts that error-recovery
behavior into a single shared helper so the pause/log policy lives in
one place. Closes #994.

## Changes

- `src/services/failure-pause-tracker.ts` — new
  `FailurePauseTracker<TState extends FailurePauseState>` that owns the
  prev-state -> bump counter -> derive `pausedDueToErrors` -> save -> log
  ladder, plus `recordSuccess` / `reset` / `isPaused`. Exports a single
  `MAX_CONSECUTIVE_FAILURES`. Pure delegation — callers wire
  `getState`/`setState` to their existing `saveState` plumbing, so each
  manager keeps its own sidecar format. Entity-specific fields
  (`lastRunAt` on success, `nextRunAt` on reset) flow through an optional
  `patch` argument.
- `src/services/hook-manager.ts` — deletes the local
  `MAX_CONSECUTIVE_FAILURES`, constructs a `FailurePauseTracker`, and
  delegates `recordSuccess` / `recordFailure` / `resetHook` to it.
- `src/services/scheduled-task-manager.ts` — deletes the local
  `MAX_CONSECUTIVE_FAILURES`, constructs a `FailurePauseTracker`, and
  delegates the inline `executeTask` success/failure blocks and
  `resetTask` to it (passing the `lastRunAt` / `nextRunAt` patches).
- `test/services/failure-pause-tracker.test.ts` — 15 new tests covering
  the ladder transitions (first-failure -> counter=1, threshold ->
  `pausedDueToErrors`, success -> reset), the per-manager log divergence
  (`logFailures`), reset-with-patch, `isPaused`, and async persistence.

Behavior is unchanged: the loop-detection path (`recordPausedDueToLoop`)
and the "is paused, skipping" early-exit log sites are left in place, and
per-manager log prefixes (`[HookManager]` / `[ScheduledTaskManager]`) are
preserved via the `label` parameter. Net -44 lines across the two
managers.

## Screenshots / Screencast

No UI changes — backend/internal refactor only. Verified live against a
test vault: forced 3 consecutive failing scheduled-task runs (bogus
model) -> `consecutiveFailures` 1 -> 2 -> 3 with `pausedDueToErrors`
flipping true at the threshold; `resetTask` cleared the failure state and
set a fresh `nextRunAt`; a successful run cleared the counter and set a
fresh `lastRunAt`.

### Required

### AI-Generated Code

- [x] AI tool(s) used: Claude (claude-opus-4-8 via Claude Code)

## Summary by CodeRabbit

* **Refactor**
  * Centralized consecutive-failure tracking and automatic pause/resume logic so both hook handling and scheduled tasks behave consistently.
  * Failure outcomes now uniformly clear on success, pause when the threshold is reached, and optionally log failure details.
* **Tests**
  * Added/expanded automated coverage for the shared failure tracker and for scheduled task execution behavior (success, failure, and cancelled runs).

### DIFF
--- a/src/services/failure-pause-tracker.ts
+++ b/src/services/failure-pause-tracker.ts
@@ -1,0 +1,135 @@
+import { getRawErrorMessage } from '../utils/error-utils';
+import type { Logger } from '../utils/logger';
+
+/**
+ * Shared "auto-pause after N consecutive failures" behavior for the markdown-defined
+ * feature managers — HookManager and ScheduledTaskManager. Both walk the identical
+ * prev-state → bump counter → derive `pausedDueToErrors` → save → log ladder over their
+ * per-entry JSON sidecar state, and both expose the same "clear the failure state" reset.
+ * This is the error-recovery layer on top of the sidecar scaffolding in `feature-definition.ts`:
+ * that file owns the state I/O, this file owns the failure/pause policy that mutates it.
+ */
+
+/** Pause an entity after this many consecutive failures. Shared by every subsystem that auto-pauses. */
+export const MAX_CONSECUTIVE_FAILURES = 3;
+
+/**
+ * The slice of per-entity sidecar state the failure-pause ladder reads and writes.
+ * Managers embed these fields in their own state shape (HookState, TaskState, …).
+ */
+export interface FailurePauseState {
+	/** Error message from the most recent failed run, if any. */
+	lastError?: string;
+	/** Number of consecutive failures since the last success. */
+	consecutiveFailures?: number;
+	/** When true the entity is auto-paused until manually reset. */
+	pausedDueToErrors?: boolean;
+}
+
+/** Outcome of {@link FailurePauseTracker.recordFailure}. */
+export interface FailureOutcome {
+	consecutiveFailures: number;
+	pausedDueToErrors: boolean;
+}
+
+export interface FailurePauseTrackerOptions<TState extends FailurePauseState> {
+	/** Read the current in-memory record for a slug (undefined if none yet). */
+	getState: (slug: string) => TState | undefined;
+	/** Persist the next record for a slug — callers wire this to `this.state[slug] = next; saveState()`. */
+	setState: (slug: string, next: TState) => void | Promise<void>;
+	logger: Logger;
+	/** Message prefix, e.g. `[HookManager]`. */
+	label: string;
+	/** Entity noun used in log messages, e.g. `Hook` or `Task`. */
+	entityNoun: string;
+	/** Failures before pausing. Defaults to {@link MAX_CONSECUTIVE_FAILURES}. */
+	maxFailures?: number;
+	/**
+	 * When true, a non-pausing failure is logged at error level (the hook manager's
+	 * behavior). When false, the caller owns failure logging — e.g. the scheduled-task
+	 * manager re-throws and lets the background runner surface the error. Defaults to false.
+	 */
+	logFailures?: boolean;
+}
+
+/**
+ * Owns the shared "auto-pause after N consecutive failures" ladder over a manager's
+ * per-entity sidecar state: bump the counter, derive `pausedDueToErrors`, persist, log.
+ *
+ * Pure delegation — the tracker never touches the state file directly. Callers wire
+ * `getState`/`setState` to their existing in-memory record + `saveState()` plumbing, so
+ * each manager keeps its own sidecar format. Entity-specific fields (a fresh `lastRunAt`
+ * on success, a fresh `nextRunAt` on reset, …) flow in through the optional `patch`
+ * argument without the tracker needing to know about them.
+ */
+export class FailurePauseTracker<TState extends FailurePauseState> {
+	private readonly maxFailures: number;
+	private readonly logFailures: boolean;
+
+	constructor(private readonly options: FailurePauseTrackerOptions<TState>) {
+		this.maxFailures = options.maxFailures ?? MAX_CONSECUTIVE_FAILURES;
+		this.logFailures = options.logFailures ?? false;
+	}
+
+	/** True when the entity is currently auto-paused. */
+	isPaused(slug: string): boolean {
+		return this.options.getState(slug)?.pausedDueToErrors ?? false;
+	}
+
+	/** Clear the failure counters after a successful run. `patch` carries entity-specific fields. */
+	async recordSuccess(slug: string, patch?: Partial<TState>): Promise<void> {
+		const prev = this.options.getState(slug);
+		const next = {
+			...(prev ?? {}),
+			...patch,
+			lastError: undefined,
+			consecutiveFailures: 0,
+			pausedDueToErrors: false,
+		} as TState;
+		await this.options.setState(slug, next);
+	}
+
+	/**
+	 * Record a failed run: bump the counter, pause once it reaches the threshold, persist,
+	 * and warn when newly paused. `patch` carries entity-specific fields.
+	 */
+	async recordFailure(slug: string, error: unknown, patch?: Partial<TState>): Promise<FailureOutcome> {
+		const prev = this.options.getState(slug);
+		const consecutiveFailures = (prev?.consecutiveFailures ?? 0) + 1;
+		const pausedDueToErrors = consecutiveFailures >= this.maxFailures;
+		const next = {
+			...(prev ?? {}),
+			...patch,
+			lastError: getRawErrorMessage(error),
+			consecutiveFailures,
+			pausedDueToErrors,
+		} as TState;
+		await this.options.setState(slug, next);
+
+		const { logger, label, entityNoun } = this.options;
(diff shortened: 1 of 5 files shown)


src/services/failure-pause-tracker.ts
```diff
@@ -0,0 +1,135 @@
...
+	/** Clear the failure counters after a successful run. `patch` carries entity-specific fields. */
+	async recordSuccess(slug: string, patch?: Partial<TState>): Promise<void> {
+		const prev = this.options.getState(slug);
+		const next = {
+			...(prev ?? {}),
+			...patch,
+			lastError: undefined,
+			consecutiveFailures: 0,
+			pausedDueToErrors: false,
+		} as TState;
```
allenhutchison: **`as TState` is an unsound (currently safe) cast worth tightening.** This same `{ ...(prev ?? {}), ...patch, ...invariants } as TState` pattern is in all three methods (here, `recordFailure` at line 106, `reset` at line 132).

`TaskState.nextRunAt` is **required** (`src/services/scheduled-task-manager.ts`). When `prev` is `undefined` and `patch` omits `nextRunAt`, this persists an object with no `nextRunAt` but casts it to a valid `TaskState` — the checker is blind to it. It never fires today only because `advanceState` always populates state before `executeTask` runs, but that's a call-ordering convention in another method, unchecked by the compiler. A future caller invoking `recordSuccess` on a fresh slug would silently store `nextRunAt: undefined`, and `new Date(undefined)` downstream becomes `Invalid Date`.

Cheapest honest fix: mirror `reset`'s existing `if (!prev) return` guard in `recordSuccess`/`recordFailure` (throw or no-op on a fresh slug) so the real precondition — "every call site has a pre-existing entry" — is enforced rather than assumed.

Optional bonus: narrow the patch type to exclude the fields the tracker owns, so the escape hatch can't even express an override the runtime silently discards (currently only guaranteed at runtime, tested at line 131):
```ts
type EntityPatch<TState> = Partial<Omit<TState, keyof FailureOutcome | 'lastError'>>;
```

allenhutchison (commented):
Thanks for this — the extraction is clean and behaviorally faithful. I verified the important equivalences (the `logFailures` asymmetry between the two managers, the scheduled-task re-throw contract, the patch-before-invariants ordering, and memory-before-disk write ordering all match the old inline code), and the new unit suite is excellent. Two non-blocking improvements before merge:

**1. Add scheduled-side integration coverage for the wiring.** The new unit tests cover `FailurePauseTracker` in isolation, and `HookManager` already drives the ladder end-to-end through `handleEvent` (`test/services/hook-manager.test.ts:602`, `:1012`, `:1103`). But the **ScheduledTaskManager success/failure paths are never exercised through the manager**: in `test/services/scheduled-task-manager.test.ts:24` the `submit` mock is `vi.fn().mockReturnValue('bg-task-1')`, which returns an ID without ever invoking the work callback. So `executeTask`'s `recordSuccess(slug, { lastRunAt })` and `recordFailure(...) + throw` paths have no manager-level coverage — a wiring mistake (wrong `getState`/`setState` closure, dropped `lastRunAt` patch, swapped `entityNoun`, missing/extra `logFailures`) would pass the whole suite. This is a pre-existing gap, but moving the logic into a config object is exactly what wiring tests protect.

Could you add two tests where the `submit` mock actually invokes the work callback: one where the runner resolves an output path (assert `lastRunAt` set + failure counters cleared), and one where it rejects (assert `consecutiveFailures` bump, `lastError`, `pausedDueToErrors` at threshold, and that the error is re-thrown)? That brings the scheduled side to parity with the hook side.

See the inline comment for the second item (type soundness of the `as TState` casts).

allenhutchison:
Thanks for the contribution.

PR #1061, opened 2026-06-29
feat(agent): parallel tool execution for independent read and search tools

yuangr:
## Summary

This PR implements parallel tool execution for independent, read-only, and search tools in the agent execution loop. This reduces latency significantly when the model decides to run multiple read or search operations in a single turn.

## Changes

- Introduced logic in \src/agent/agent-loop.ts\ to identify and separate tool calls into parallelizable and serial groups.
- Read-only tools (\ToolClassification.READ\) and search/external tools (\ToolClassification.EXTERNAL\) that do not require user confirmation are executed concurrently using \Promise.all\.
- State-modifying write/delete tools and tools requiring user confirmation continue to execute sequentially to prevent race conditions and preserve user control.
- Added comprehensive unit tests in \	est/agent/agent-loop.test.ts\ to verify the execution order and concurrent startup of parallel tools.

### Required

- [ ] This PR is linked to an approved issue where the approach was discussed with a maintainer (N/A - Opened directly for performance optimization review)
- [x] All CI checks pass (\
pm test\, \
pm run build\, \
pm run format-check\)
- [ ] Documentation has been updated (if applicable) (N/A - Internal performance optimization with no settings changes)

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Tool calls are now scheduled more intelligently: eligible read/external actions may run in parallel, while other actions continue to run sequentially for safer execution.
* **Bug Fixes**
  * Improved cancellation behavior so cancelled batches and in-flight tools return consistent “Cancelled” results promptly and more reliably.
* **Tests**
  * Added coverage to verify parallel vs sequential start timing based on tool classifications.

### DIFF
--- a/src/agent/agent-loop.ts
+++ b/src/agent/agent-loop.ts
@@ -619,7 +619,11 @@ export class AgentLoop {
 
 		if (isCancelled()) {
 			plugin.logger.debug('[AgentLoop] Cancellation detected before tool batch execution');
-			return results;
+			return sortedToolCalls.map((toolCall) => ({
+				toolName: toolCall.name,
+				toolArguments: toolCall.arguments || {},
+				result: { success: false, error: 'Cancelled' },
+			}));
 		}
 
 		// Split tool calls into parallelizable and serial (confirmation-requiring or write/destructive)
@@ -725,10 +729,16 @@ export class AgentLoop {
 		}
 
 		// Execute serial calls sequentially
+		let serialCancelled = false;
 		for (const toolCall of serialCalls) {
-			if (isCancelled()) {
-				plugin.logger.debug('[AgentLoop] Cancellation detected, stopping serial tool execution');
-				break;
+			if (serialCancelled || isCancelled()) {
+				serialCancelled = true;
+				results.push({
+					toolName: toolCall.name,
+					toolArguments: toolCall.arguments || {},
+					result: { success: false, error: 'Cancelled' },
+				});
+				continue;
 			}
 
 			const executionId = `${toolCall.name}-${Date.now()}-${Math.random().toString(36).substring(2, 11)}`;
--- a/src/agent/agent-loop.ts
+++ b/src/agent/agent-loop.ts
@@ -1,7 +1,7 @@
 import type { Content } from '@google/genai';
 import type ObsidianGemini from '../main';
 import type { ChatSession, PerTurnContext } from '../types/agent';
-import type { FeatureToolPolicy } from '../types/tool-policy';
+import { ToolClassification, type FeatureToolPolicy } from '../types/tool-policy';
 import type { ToolCall, ModelResponse, ModelApi, StreamChunk } from '../api/interfaces/model-api';
 import type { CustomPrompt } from '../prompts/types';
 import type { IConfirmationProvider, IToolHostView, ToolExecutionContext, ToolResult } from '../tools/types';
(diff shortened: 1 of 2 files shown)


allenhutchison:
## Review

Thanks for this — the parallel/serial split is a well-designed approach, and the confirmation gate is correctly mirrored from the execution engine, so a tool that would prompt can't be misclassified into the parallel bucket. Tests pass (42/42) and `typecheck:test` is clean. A few things to address before merge, then some polish and test gaps.

### Result-ordering concern — investigated, it's safe (no code change needed)

I dug into whether reordering results as `[...parallel, ...serial]` could mis-pair responses to calls (especially duplicate tool names). It does **not** introduce a correctness regression:

- `buildToolHistoryTurns` already pairs the `functionCall` turn (built from unsorted `currentToolCalls`) with the `functionResponse` turn (reordered results) on master too, since `sortToolCallsByPriority` already hoists reads above writes. Pairing is **by name**, not position (`functionResponse.id` is never set; the interactions-mapper matches by name).
- Two calls to the **same tool name** can't straddle the split — confirmation status is computed per-name, so same name ⇒ same classification ⇒ same bucket. `Promise.all` preserves input order and both buckets are built by stable iteration over `sortedToolCalls`, so same-name relative order is preserved exactly as on master.

→ Worth a one-line code comment + a regression test (two same-name reads) to lock this invariant in, since a future switch to `allSettled`-with-sort, or splitting same-name calls across buckets, would break it silently.

### Important — resilience regressions introduced by this PR

**1. Unguarded partition loop can abort the whole batch** — `src/agent/agent-loop.ts:633-654`
The new classification loop calls `requiresConfirmation(...)` and `confirmationProvider.isToolAllowedWithoutConfirmation(...)` outside any try/catch. On master, every tool ran inside a per-tool try that converted a throw into `{success:false}`. Now a single throw here (e.g. a malformed `featureToolPolicy` or a throwing confirmation provider) rejects all of `executeToolBatch` and discards results for *every* tool in the batch. Please wrap the loop body and default an un-classifiable call to `serialCalls` (the safe path that still gets per-tool error handling).

**2. Description builder moved outside the try (parallel path)** — `src/agent/agent-loop.ts:673-675`
In the parallel closure, `getProgressDescription`/`generateToolDescription` run *before* the `try` at line 682, so a throwing description builder rejects that promise → rejects `Promise.all` → aborts the batch. On master this ran inside the try, and the serial path here (746-751) still correctly keeps it inside. Live risk is low (current impls are null-guarded) but the safety envelope is gone — please move the description computation inside the try.

The cleanest fix for both is to extract the shared per-tool execute-and-emit logic into a private `runSingleTool(toolCall, toolContext, options)` helper used by both paths (see Suggestions).

### Suggestions

- **De-duplicate the two ~50-line execution bodies** into a private `runSingleTool(...)` helper. This removes drift risk between the parallel and serial copies (the lost comment below is an early symptom), fixes #2 in one place, and normalizes the `toolArguments` inconsistency.
- **Optional concurrency cap (e.g. 4-6)** on the parallel batch. Low severity — EXTERNAL tools (`web_fetch`, `google_search`, `deep_research`) map to `ASK_USER` under the default preset and so fall into the serial path; fanout only happens under EDIT_MODE/YOLO or explicit auto-approve. Worst case is several `deep_research` firing at once under YOLO.
- **Restore the deleted rationale comment** on the serial-path `logger.warn` ("the engine's early-return paths return `{success:false}` without logging themselves") — the log itself remains in both paths, only the explanation was lost.
- **`toolArguments` consistency** — parallel success returns `toolCall.arguments`; the cancel/error paths use `toolCall.arguments || {}`. Align them.
- **Docs (AGENTS.md):** per the repo's mandatory-docs rule, the `AgentLoop` description (architecture item 6, and ideally pattern #9 "Tool Execution Order") should note that read/external tools now run concurrently while write/destructive/confirmation-requiring tools stay serial.

### Test gaps (the change has one timing-based test today)

- **Confirmation-required READ stays serial** — the core safety gate, currently untested since the test hardcodes `requiresConfirmation: () => false`. Highest-value addition: a READ tool whose `requiresConfirmation()` is true must land in the serial bucket.
- **EXTERNAL classification** parallelization — the test only covers READ vs WRITE, never EXTERNAL.
- **Missing-tool** (`getTool` returns undefined) → routes to the serial path.
- **Cancellation paths** — pre-batch early-return, per-parallel-promise check, and the serial `serialCancelled` short-circuit; assert the returned pairs still cover every input call.
- **Individual parallel failure** — one parallel tool throws/rejects, the others still resolve with results and the batch isn't rejected.
- **Flaky timing test** (`test/agent/agent-loop.test.ts:1359-1421`): real `setTimeout(50)` with `<20ms` / `-5ms` wall-clock assertions is fragile under CI load. Consider converting to controllable deferred promises that assert invocation ordering deterministically (both reads invoked before either resolves; write not invoked until both reads resolve) — that tests the same contract without wall-clock races and naturally covers the ordering invariant and the partial-failure case.

### Strengths
- Confirmation gate exactly mirrors `execution-engine.ts:109-116`, and routing confirmation tools to serial correctly avoids racing confirmation modals.
- Loop-detection accounting is unaffected (check-then-record is synchronous in the engine; per-turn abort counting happens after the batch).
- Defensive coding (`typeof requiresConfirmation === 'function'`, optional-chained tool re-fetch) keeps the new path resilient.

allenhutchison:
Closed due to lack of response. Please reopen if you plan to continue working on this.

PR #1062, opened 2026-06-29
feat(api): Google Context Caching for large agent sessions

yuangr:
## Summary

This PR integrates the Google Gemini Context Caching API to cache conversation history prefixes for long chat and agent sessions. This reduces token usage costs and response latency for sessions exceeding 32,768 tokens.

## Changes

- Implemented TTL-based prefix caching in \src/api/providers/gemini/client.ts\ using \caches.create\.
- Cached prefixes are reused for subsequent turns within the session as long as the history prefix remains stable.
- Added a \contextCachingEnabled\ toggle setting in \src/main.ts\ and \src/ui/settings-agent-config.ts\ (enabled by default).
- Added settings label and description in English (\src/i18n/en.ts\).
- Added comprehensive unit tests in \	est/api/providers/gemini/client.test.ts\ to assert that cache resources are created, reused, and expired correctly, and that cached request payloads are correctly formatted.

### Required

- [ ] This PR is linked to an approved issue where the approach was discussed with a maintainer (N/A - Opened directly for review of cost-optimization features)
- [x] All CI checks pass (\
pm test\, \
pm run build\, \
pm run format-check\)
- [ ] Documentation has been updated (if applicable) (N/A - Internal optimization, toggle documented in inline descriptions)

### AI-Generated Code

### DIFF
--- a/src/api/providers/gemini/client.ts
+++ b/src/api/providers/gemini/client.ts
@@ -86,6 +86,18 @@ export class GeminiClient implements ModelApi {
 	private prompts: GeminiPrompts;
 	private plugin?: ObsidianGemini;
 
+	private static activeCaches = new Map<
+		string,
+		{
+			cacheName: string;
+			model: string;
+			systemInstruction: string;
+			toolsJson: string;
+			cachedTurnsJson: string;
+			expiresAt: number;
+		}
+	>();
+
 	constructor(config: GeminiClientConfig, prompts?: GeminiPrompts, plugin?: ObsidianGemini) {
 		this.config = {
 			temperature: 1.0,
@@ -441,9 +453,6 @@ export class GeminiClient implements ModelApi {
 			systemInstruction = request.prompt || '';
 		}
 
-		// Build conversation contents
-		const contents = this.buildContents(request);
-
 		// Build config
 		const config: any = {
 			temperature: request.temperature ?? this.config.temperature,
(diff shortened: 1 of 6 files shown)


allenhutchison:
## Review

Thank you for the work here — and first, a process note that matters more than any single line of code below.

### Let's decide whether we want this feature at all — before more code

The core question this PR raises isn't "is the implementation correct?" — it's **"does explicit context caching buy us anything over the implicit caching Gemini already does for us?"** My assessment is **no**, for this plugin's usage, and I don't think we should merge it. Details below, but I want to be upfront that this is a "should we build it" decision, not a "fix these bugs" review.

This is exactly why our [Contributing Guidelines](../CONTRIBUTING.md) ask for an **approved issue where the approach is discussed with a maintainer before code is written**. The checklist item "This PR is linked to an approved issue" is unchecked here (marked N/A), and I think that's the root of the wasted effort: a quick issue discussion up front would have surfaced the points below and saved you from implementing ~260 lines that I'm going to recommend we not take. Please do open an issue first for the next idea — I'm very happy to talk through design there early. This isn't a knock on the work; it's the part of the process that protects your time.

### Why explicit caching doesn't buy us anything here

1. **It never runs in production as wired.** The caching block is gated on `config.sessionId`, but the production factory (`ModelClientFactory.createFromPlugin`, `src/api/factory.ts`) never sets `sessionId` — only the tests inject it. So the guard `if (isExtended && sessionId && contextCachingEnabled)` (`client.ts:500`) is always false, `caches.create` is never called, and the default-on setting controls nothing. The passing tests give false confidence because they construct the client with a `sessionId` no real caller supplies.
2. **Even fully wired, it duplicates implicit caching — and worse.** Per Google's docs, implicit caching is **on by default for all Gemini 2.5+ models**, requires zero code, and automatically passes through the cache-hit discount. Its threshold is **2,048 tokens (2.5) / 4,096 (3.x)**; this PR only engages at **32,768**. Below 32k implicit already caches and this code is idle; above 32k implicit *also* already caches the stable prefix. We'd be reimplementing — with a cruder `json.length/4` token estimate and manual O(n) prefix matching — what the platform does for free at a far lower threshold.
3. **It's on the path we're migrating away from.** Explicit `caches.create` lives only in the legacy `generateContent` path and is **not supported in the Interactions API**, which the plugin is actively moving to (`useInteractionsApi`). So even if `sessionId` were plumbed, the feature goes inert as Interactions becomes the default.
4. **Cost can go the wrong way.** Explicit caches bill for *storage* over the TTL regardless of whether you get a second hit. With a 5-minute TTL on interactive sessions where a user often reads one answer and stops, we'd pay storage for caches that are never reused — potentially *raising* cost vs implicit, which charges nothing for storage and only ever discounts real hits. Explicit caching pays off for large, stable context reused across many requests inside the TTL (batch / high-QPS over a fixed corpus) — not interactive chat.

### One thing that must be addressed regardless of the above

Independent of the (dead) caching feature, this PR ships a **live regression to every Gemini request** on the legacy path:

- **`buildContents` now silently drops history entries** — `src/api/providers/gemini/client.ts:638`. The guard changed from `'parts' in entry` to `entry.parts?.length`, so any entry with empty/missing `parts` (e.g. a model turn carrying only function calls) is dropped instead of passed through, and each entry is reconstructed as `{role, parts}`, discarding other fields. This is not gated by `sessionId`, so it affects all users today.
- It also **diverges from the Interactions path** (`client.ts:289` still uses `'parts' in entry`), so identical input yields different model context depending on transport.

If anything from this PR lands, it should only be this line — reverted, or split into its own correctly-scoped PR with a regression test. As written it's a behavior change with no coverage.

### If the feature were ever pursued (notes for the record)

These don't need fixing given the recommendation, but for completeness: the `static activeCaches` map is shared across all per-use-case client instances and keyed only by `sessionId` (cross-use-case collisions evict still-billing caches; no lifecycle cleanup or size cap); the `json.length / 4` token estimate counts JSON envelope overhead and can pass the local 32k gate then get rejected by the API; and the check-then-create sequence isn't atomic, so parallel same-session requests both create caches and orphan one for its full billed TTL. The new setting also lacks the documentation our guidelines require for user-facing changes.

### Recommendation

I'd suggest we **close or substantially narrow this PR**. Implicit caching already gives us the cost/latency benefit for this plugin's usage with no code to maintain. The only change worth salvaging is the `buildContents` fix — and only if it's reverted or re-scoped with a test. For future work, please open an issue first so we can make the build/don't-build call together before you invest the time. Genuinely appreciate the effort and the thoroughness of the tests — I'd love to see it pointed at something we've agreed to ship.

_(Note: I found that running the suite required a clean working tree — stale local build artifacts can shadow the `.ts` sources and make these tests appear to fail; on a clean tree all 84 pass. Not a PR issue.)_

allenhutchison:
Closed due to lack of response. Please reopen if you plan to continue working on this.

PR #1082, opened 2026-07-02
feat(attachments): rasterize SVG/SVGZ to PNG before inlining (#641)

### LINKED ISSUE #641: Feat: Add support for SVG images
I started using the [Obsidian Ink](https://github.com/daledesilva/obsidian_ink) plugin via BRAT to use version 0.4.0 on my Pixel tablet to support handwritten text via stylus. The 0.4.0 version auto-converts the strokes to SVG and embeds it in the markdown file as an attachment. Using the Gemini Chat window, I asked the agent to convert the image to text and insert it into the note. It tried really hard, found the file in the listing, but was seemingly unable to read it. Eventually it responded with the following:

> I've located the ink writing image in your `Untitled.md` file , which points to the file `Ink/Writing/2026.4.15 - 11.31am.svg`.
>
> Unfortunately, I am currently unable to read SVG files directly from your vault to perform OCR (Optical Character Recognition) on the handwriting. I can process standard image formats like JPG or PNG, but the SVG format used by th eink plugin isn't supported for direct viewing in my current toolset.

It goes on to suggest I take a screenshot and paste it into the chat. 

Well, Gemini for web definitely supports SVGs and Nano Banana can even generate SVGs. I was able to attach it to a web chat prompt and it converted the image without issue. Presumably, this just requires adding a mime type definition mapping `.svg` (and technically `.svgz`) to `image/svg+xml`. It's not technically a binary format, but adding it to these two lists might make it work?

https://github.com/allenhutchison/obsidian-gemini/blob/82d6f1e90285d8ab07abbecb2ef6279988a78c55/src/ui/agent-view/inline-attachment.ts#L70

https://github.com/allenhutchison/obsidian-gemini/blob/82d6f1e90285d8ab07abbecb2ef6279988a78c55/src/utils/file-classification.ts#L17

allenhutchison:
## Summary

Gemini's inline-data endpoint rejects `image/svg+xml` outright, so SVG files could never be sent to the model — the reported symptom was the agent locating a vault `.svg` (Obsidian Ink handwriting export) but being unable to read it. This adds **client-side SVG → PNG rasterization** at the shared classify/inline boundary, so `.svg` and `.svgz` files are rendered to PNG (via an off-screen Electron `<canvas>`) before hitting the inline-base64 path. Rasterizing — rather than sending raw XML as text — is the only path that lets the model view and OCR ink `<path>` strokes.

Produced by the auto-dev pipeline from the approved plan on #641.

Fixes #641

## Changes

- **`src/utils/svg-rasterizer.ts` (new)** — `rasterizeSvg(buffer, isSvgz)`: decompresses `.svgz` via `DecompressionStream('gzip')`, loads the SVG into an ``, and draws it onto an off-screen `<canvas>` scaled to at most **2048px on the longest edge** (aspect-ratio preserved) over a **white background** (transparent ink would otherwise flatten to black and hurt OCR). Returns a base64 PNG; throws on malformed/unloadable SVG so callers can fall back. Exports `computeScaledDimensions`/`isSvgExtension` helpers.
- **`src/utils/file-classification.ts`** — new `FileCategory.SVG` returned for `svg`/`svgz` (mimeType `image/svg+xml`).
- **Wired SVG handling into every inline entry point** (one rasterization path, not four): vault drag-drop and external drop/paste (`agent-view-ui.ts`), the `@`-mention picker (`agent-view-attachments.ts`), and the **Read File tool** (`read-file-tool.ts`, the actual reported scenario). On rasterization failure each path falls back to the existing "unsupported file type" notice / an error string rather than sending anything unusable.
- **`file-mention-modal.ts`** — SVG files show the image icon and remain selectable in the picker.
- **Tests** — `test/utils/svg-rasterizer.test.ts` (scaling/cap, white-background fill, `.svgz` decompression, malformed-SVG rejection) and new SVG cases in `file-classification.test.ts`.
- **Docs** — `docs/guide/agent-mode.md` (routing table, supported-formats list, new "SVG Handling" section, Read File examples) and `README.md`.

**Deviation from the plan:** the plan listed the `@`-mention edit under `agent-view.ts`; that flow actually lives in `agent-view-attachments.ts`, and external drop/paste of raw SVG `File` objects is handled in `agent-view-ui.ts`. Scope, sizing, white background, `.svgz`, and fallback behavior all follow the spec.

## Screenshots / Screencast

N/A — no visual UI change; SVG attachments render as standard PNG image thumbnails through the existing attachment preview.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#641)
- [ ] I have verified this change does not break Mobile — rasterization relies only on `<canvas>`, `Image`, and `DecompressionStream`, all available in Obsidian's mobile WebView, but this was **not** manually verified on a mobile device

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01WtTRvGn5fUV73Ez7BXYytV)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Let the AI actively work with your vault through tool calling capabilities.
 
 - **Search Files by Name:** Find any file by filename patterns (wildcards supported)
 - **Search File Contents:** Grep-style text search within note contents (supports regex and case-sensitive search)
-- **Read Files:** Access text files or analyze binary files (images, audio, video, PDF) directly through Gemini
+- **Read Files:** Access text files or analyze binary files (images, audio, video, PDF, SVG) directly through Gemini — SVGs are rasterized to PNG on-device so vector artwork and handwritten ink can be viewed and OCR'd
 - **Create Notes:** Generate new notes with specified content
 - **Edit Notes:** Modify existing notes with precision
 - **Move/Rename Files:** Reorganize and rename notes in your vault
--- a/docs/guide/agent-mode.md
+++ b/docs/guide/agent-mode.md
@@ -72,7 +72,7 @@ Each tool call in the chat is collapsible — click a tool row to expand its det
 
 ### File Attachments & Drag-and-Drop
 
-You can include images, audio, video, PDFs, and text files in your chat. Files are automatically classified and routed:
+You can include images, audio, video, PDFs, SVGs, and text files in your chat. Files are automatically classified and routed:
 
 **Adding Files:**
 
@@ -90,14 +90,20 @@ When you drop a file, the plugin classifies it based on its extension:
 | ----------------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------------------ |
 | **Text**                      | `.md`, `.txt`, `.ts`, `.js`, `.json`, `.html`, `.css`, `.py`, etc.            | Added as context chips (the AI reads the file content)             |
 | **Binary (Gemini-supported)** | `.png`, `.jpg`, `.gif`, `.webp`, `.pdf`, `.mp3`, `.wav`, `.mp4`, `.mov`, etc. | Sent as inline data (the AI processes the binary content directly) |
+| **SVG**                       | `.svg`, `.svgz`                                                               | Rasterized to PNG on-device, then sent as inline data              |
 | **Unsupported**               | `.zip`, `.exe`, `.dmg`, etc.                                                  | Skipped with a notification                                        |
 
 **Supported Binary Formats:**
 
 - **Images**: PNG, JPEG, GIF, WebP, HEIC, HEIF
+- **Vector images**: SVG, SVGZ (rasterized to PNG before sending — see below)
 - **Audio**: WAV, MP3, AAC, FLAC
 - **Video**: MP4, MPEG, MOV, FLV, MPG, WebM, WMV, 3GP
 - **Documents**: PDF
+
+**SVG Handling:**
+
+Gemini's API can't process `image/svg+xml` directly, so SVG (and gzip-compressed `.svgz`) files are **rasterized to PNG on your device** before being sent — whether you drag, paste, `@`-mention, or have the agent read them with the Read File tool. Rasterization renders the SVG onto a white background (so transparent artwork like handwritten ink strokes stays legible for OCR) and caps the longest edge at 2048px to keep the payload within the inline-data limit. If an SVG can't be rendered (malformed markup or unresolvable external references), it's skipped with the same "unsupported file type" notice rather than sending unusable data.
 
 **How It Works:**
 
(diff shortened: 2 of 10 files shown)


allenhutchison:
Reviewed and tested this end-to-end in a real vault (drove `rasterizeSvg` through the `read_file` tool and decoded the resulting PNG bytes). It works well — normal SVGs, the 2048px cap (5000×3000 → 2048×1229, aspect preserved), `.svgz` gzip decompression, and malformed-SVG rejection all behave correctly. Nice work. One actionable item and two FYIs before merge:

**1. (actionable) viewBox-only SVGs rasterize at a reduced intrinsic size, not the viewBox size.**
In `src/utils/svg-rasterizer.ts`, the `parseSvgDimensions` fallback only runs when the loaded `` reports `naturalWidth`/`naturalHeight` of `0`. But for an SVG that has a `viewBox` and no explicit `width`/`height`, Chromium reports a *nonzero default* intrinsic size rather than 0 — e.g. a `viewBox="0 0 400 300"` SVG comes back as `200×150`, so the fallback never fires and the PNG is rasterized at half resolution (correct aspect ratio, just smaller/blurrier).

Suggested fix: when the SVG markup declares a `viewBox` but no explicit `width`/`height`, prefer the viewBox's width/height as the render size instead of the img's default intrinsic size (fall back to the intrinsic size only when there's no usable viewBox). The reported use case — Obsidian Ink handwriting — exports explicit `width`/`height`, so this is a lower-priority polish, not a blocker, but worth fixing since diagram/icon SVGs are commonly viewBox-only. Please add a test with a viewBox-only SVG asserting the rasterized dimensions match the viewBox.

**2. (trivial, optional) import-style inconsistency.** `agent-view-ui.ts` imports `rasterizeSvg` statically while `agent-view-attachments.ts` uses a dynamic `await import(...)`. Harmless, but worth making consistent (a static import is fine in both).

**3. (FYI, no action) mobile.** The checklist notes mobile wasn't verified on-device. `DecompressionStream`/`<canvas>`/`Image` are all present in modern mobile WebViews, so this is almost certainly fine — just flagging that it's untested.

Everything else — the four wired entry points, the fallbacks, the white-background rationale, and the security posture (loading SVG via `` disables scripts) — looks correct. Address #1 (and #2 if easy) and this is good to merge.

allenhutchison:
Thanks for the thorough end-to-end test. Addressed items 1 and 2 in `2c8df0e`:

**1. viewBox-only SVGs (fixed).** You're right — the `naturalWidth === 0` guard never fired for viewBox-only SVGs because Chromium hands back a reduced default intrinsic size. Reworked the sizing to resolve from the markup first: **explicit `width`/`height` → viewBox extent → browser intrinsic size → default**. Percentage `width`/`height` are now treated as unusable (0) so the viewBox is preferred in that case too. Added a viewBox-only test (`viewBox="0 0 400 300"` with a mocked 200×150 intrinsic size) asserting the canvas rasterizes at 400×300, not 200×150.

**2. import consistency (done).** `rasterizeSvg` is now a static top-level import in `agent-view-attachments.ts`, matching `agent-view-ui.ts`.

**3. mobile (noted, no change).** Agreed it's almost certainly fine given `DecompressionStream`/`<canvas>`/`Image` are all in modern mobile WebViews; leaving the checklist item honest as untested-on-device.

Full pre-flight green after the change: build, `npm test` (3209 passing), format-check, and test typecheck.

---

PR #1084, opened 2026-07-02
fix(history): stop tool logs folding into the preceding reasoning callout

### LINKED ISSUE #1050: Agent session files tools are getting subsumed into reasoning blocks
The Agent Session File is supposed to have alternating callouts for reasoning and tool use:

Reasoning: reasoning text in a call out

Tool Execution: Tool call in a call out

Reasoning: more reasoning text in a call out

Tool Execution: More tool calls

However this is what I'm seeing right now at HEAD:

> [!reasoning]- Reasoning
> **My Investigation into AB 2047**
> 
> Okay, so the internal search in the vault came up empty. No meaningful hits for "AB 2047" – just a random "47." and a repetitive "Okay." in some old notes. The session recall tool gave me nothing either. That's a bit perplexing. Since the vault doesn't seem to hold any direct information on it, my next step is clear: I need to conduct an external search to get current, accurate information about AB 2047 as of 2026.
> 
> First, I think a simple Google search is the way to go to quickly get an overview of what AB 2047 is all about. I'll need to figure out the context of AB 2047, especially if it relates to California or any other specific area. Let's see if there's any prominent bill by that designation.
> 
> Wait a second, there's a "summary" output from the semantic search tool, which suggests the tool incorporated some general knowledge, potentially web-based. But before I rely on it, I need to verify its accuracy with a proper Google search. I'm going to conduct a more specific search, `AB 2047` and `California AB 2047` to see what comes up.
> 
> Actually, before I get too far, I need to confirm the originating state or country. "AB" usually points to a California Assembly Bill, but it's best to be certain. A search for `AB 2047 bill` should help me identify the bill's jurisdiction and uncover its status and details quickly.
> 
> 
>
> 🔧 `google_search` query="AB 2047 bill" → success (28718ms)

It looks like we are dropping the "Tool Execution" callout reference in some cases. We need to fix this.

allenhutchison:
## Summary

Fixes #1050 — tool-execution log lines were being rendered *inside* the preceding `[!reasoning]-` callout in agent session files, instead of as their own `[!tools]-` callout, so the "Tool Execution" reference was dropped and reasoning/tool activity no longer alternated.

Root cause is entirely in `mergeToolBlock` (`src/subscribers/tool-execution-logger.ts`). Its backward scan for a trailing `> [!tools]- Tool Execution` header walked straight over the lines of an intervening `[!reasoning]-` callout — every one starts with `>`, so nothing stopped it. When an earlier tool batch existed above a reasoning-only turn, the scan matched that **stale** header and spliced the new `🔧` line onto `trimmed` (the end of the file), landing it inside the trailing reasoning callout with no `[!tools]` header of its own. (The plan on the issue guessed at `session-history.ts` / `agent-loop-helpers.ts`; the actual defect was in the tool logger's merge.)

## Changes

- `src/subscribers/tool-execution-logger.ts` — the backward scan now stops at the **first** callout header it meets. If that header is not the tools header (e.g. `[!reasoning]-`), the tail block isn't a tools callout, so the new lines start a fresh, blank-line-separated `[!tools]-` block instead of merging into the preceding callout. Merging into a genuinely-trailing tools callout (consecutive tool batches) is unchanged.
- `test/subscribers/tool-execution-logger.test.ts` — two regression tests: the multi-callout fold case from the issue (prior `[!tools]` batch + trailing `[!reasoning]` turn), and a reasoning-tail blank-line separation check.

## Why the fix is minimal

The append path (`content + '\n' + block + '\n'`) is unchanged: history entries always end with a trailing newline (the reasoning callout template emits a blank line), so once the scan correctly returns "no trailing tools callout," the existing append already produces the required blank-line separation. Both existing `mergeToolBlock` test suites (`test/subscribers/…` and `test/services/…`, which pin the exact append string) stay green.

## Screenshots / Screencast

N/A — fixes the serialized markdown structure of session-history files; no UI surface changed. The behavioral effect is that a saved session again shows alternating `[!reasoning]-` and `[!tools]-` callouts (the issue has the before-state).

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — Fixes #1050 (plan approved; approach adjusted after locating the real root cause, noted above)
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`) — 3196 tests pass, build + typecheck:test + format-check clean
- [x] I have tested this change on Desktop — N/A: covered by unit tests on the pure `mergeToolBlock` function; not run interactively
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — no platform-specific code touched
- [x] Documentation has been updated (if applicable) — N/A: restores the documented session-file format (parser invariant in AGENTS.md), no behavior users configure

### AI-Generated Code

---
https://claude.ai/code/session_01Dp5HWzyqpRBjrUce8FRATF

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merging of tool-related callouts to ensure the existing tools callout is only extended when it is the trailing block, with correct spacing and separation from adjacent callouts.
  * Ensured a fresh trailing tools callout is created when appropriate to prevent unintended callout folding in Obsidian.
* **Tests**
  * Expanded unit test coverage for tool rendering to include `google_search` query and `fetch_url` URL arguments.
  * Added regression cases for merging behavior when no existing tools callout is present and when reasoning callouts precede tools.


allenhutchison:
Addressed both CodeRabbit nitpicks in 6fed9fa:

1. **Fragile trailing-newline assumption** — the fresh-block append now builds from `trimmed` with an explicit single blank line (`trimmed + '\n\n' + block`), so the separator no longer depends on `content` ending in a newline. Empty content still returns `'\n' + block + '\n'`.
2. **Duplicated header literal** — extracted `TOOLS_CALLOUT_HEADER` as the single source of truth for `formatToolBlock`, the trailing-callout scan, and the filter. Since the anchored regex only ever matched a full header line, the scan is now a plain string equality against that constant (no dynamic regex needed).

Updated the one test in `test/services/tool-execution-logger.test.ts` that pinned the old single-newline separator. Full suite green (3196), and re-verified the #1050 repro in the test vault still produces a separate `[!tools]` callout after the reasoning callout.

allenhutchison:
Addressed the location nitpick in 3fb88b1: consolidated the two same-named tool-execution-logger suites into the canonical `test/subscribers/tool-execution-logger.test.ts` (mirroring `src/subscribers/`) and removed the misplaced `test/services/` copy.

Before deleting, I ported the cases that were unique to the services suite so no coverage is lost: the `google_search` (`query=`) and `fetch_url` (`url=`) key-param mappings, merging into a header-only `[!tools]` callout, and multi-line merges. Full suite green (3180).

PR #1093, opened 2026-07-03
refactor: use vault.configDir for system-folder exclusions (#1031)

### LINKED ISSUE #1031: Replace hardcoded '.obsidian' with Vault.configDir in system-folder exclusions
Deferred from the obsidianmd dashboard cleanup (the `obsidianmd/hardcoded-config-path` rule, disabled in `eslint.config.mjs`).

Several system-folder exclusion checks hardcode the literal `.obsidian` instead of `app.vault.configDir`. A user can rename their config directory, so these checks can both miss the real config dir and over-match a folder literally named `.obsidian`.

### Sites
- `src/utils/file-utils.ts` (`shouldExcludePath` — the central helper)
- `src/services/deep-research.ts` (`PROTECTED_FOLDER_SEGMENTS`)
- `src/services/hook-manager.ts`
- `src/services/image-generation.ts`
- `src/services/obsidian-file-adapter.ts`
- `src/services/vault-analyzer.ts` (two spots)
- `src/tools/vault/read-file-tool.ts`
- `src/ui/settings-rag.ts`

### Approach
- Thread `configDir` into `shouldExcludePath` (e.g. a `configDir` param supplied by `shouldExcludePathForPlugin` from `plugin.app.vault.configDir`), and route the inline checks through the helper where possible.
- Keep matching the literal `.obsidian` as a fallback only if needed for safety.
- Add tests for a custom `configDir`.
- Re-enable `obsidianmd/hardcoded-config-path` once the code sites are fixed (the rule note flags ~35 total; some are in comments/docs that eslint won't flag).

Part of the broader obsidianmd dashboard-warning cleanup; the high-value trashFile fix landed separately.

allenhutchison:
## Summary

Several system-folder exclusion checks hardcoded the literal `.obsidian` instead of the vault's actual configuration directory (`app.vault.configDir`). Because a user can rename their config directory, these checks could both **miss the real config dir** (leaking config files into agent/RAG operations) and **over-match** a vault folder that happens to be named `.obsidian`. This threads the vault's real `configDir` through the exclusion logic and re-enables the ESLint rule that guards against the anti-pattern.

Produced by the auto-dev pipeline from the approved plan on #1031.

Fixes #1031

## Changes

- **`src/utils/file-utils.ts`** — `shouldExcludePath()` gains an optional `configDir` param; `shouldExcludePathForPlugin()` now passes `plugin.app.vault.configDir`, so every caller of the plugin wrapper (all vault tools + UI file pickers) is fixed at once. A documented `DEFAULT_CONFIG_DIR = '.obsidian'` fallback is kept for callers that don't supply a `configDir` (backward-compatible with existing behavior and `createFileFilter`).
- **Inline `.obsidian` checks routed through the vault's `configDir`:** `src/services/deep-research.ts`, `src/services/hook-manager.ts`, `src/services/image-generation.ts`, `src/services/obsidian-file-adapter.ts`, `src/services/vault-analyzer.ts` (two spots), `src/tools/vault/read-file-tool.ts`, `src/ui/settings-rag.ts`, and `src/tools/execution-engine.ts`.
- **`eslint.config.mjs`** — removed `obsidianmd/hardcoded-config-path` from the disabled `PERVASIVE_OBSIDIANMD_RULES_TODO` block so it is enforced for `src/`; scoped it off for the `test/**` glob, where fixtures legitimately use concrete `.obsidian` sample paths. The single documented fallback constant in `file-utils.ts` carries a justified inline disable directive.
- **Tests** — added `shouldExcludePath` / `shouldExcludePathForPlugin` cases for a renamed config directory (`_obsidian` excluded; a literal `.obsidian` folder not over-matched); updated service test mocks to expose `vault.configDir`.
- **Docs** — clarified in `docs/guide/semantic-search.md` that the Obsidian configuration directory is always excluded whether or not it has been renamed from `.obsidian`.

## Screenshots / Screencast

N/A — internal correctness change, no UI surface.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — approved plan on #1031
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`) — plus `npm run lint` (0 errors, `hardcoded-config-path` now enforced) and `npm run typecheck:test`
- [x] I have tested this change on Desktop — full unit suite (3225 tests) passes
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — pure path-string logic reading `vault.configDir`, no platform-specific APIs
- [x] Documentation has been updated (if applicable) — semantic-search guide clarified

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_019So7uyw4QoDkhSHyfpNmpT)_

### DIFF
--- a/src/services/hook-manager.ts
+++ b/src/services/hook-manager.ts
@@ -794,7 +794,8 @@ export class HookManager {
 		const stateFolder = normalizePath(this.plugin.settings.historyFolder) + '/';
 		if (filePath === stateFolder.slice(0, -1)) return true;
 		if (filePath.startsWith(stateFolder)) return true;
-		if (filePath.startsWith('.obsidian/')) return true;
+		const configDir = this.plugin.app.vault.configDir;
+		if (filePath === configDir || filePath.startsWith(configDir + '/')) return true;
 		return false;
 	}
 
--- a/src/services/hook-manager.ts
+++ b/src/services/hook-manager.ts
@@ -1,6 +1,6 @@
 import { Platform, TAbstractFile, TFile, normalizePath } from 'obsidian';
 import type ObsidianGemini from '../main';
-import { ensureFolderExists } from '../utils/file-utils';
+import { ensureFolderExists, shouldExcludePath } from '../utils/file-utils';
 import { FeatureToolPolicy } from '../types/tool-policy';
 import { formatToolPolicyYaml } from './feature-policy-yaml';
 import {
@@ -791,12 +791,13 @@ export class HookManager {
 	// ── Filter / gate helpers ────────────────────────────────────────────────
 
 	private isExcludedPath(filePath: string): boolean {
-		const stateFolder = normalizePath(this.plugin.settings.historyFolder) + '/';
-		if (filePath === stateFolder.slice(0, -1)) return true;
-		if (filePath.startsWith(stateFolder)) return true;
-		const configDir = this.plugin.app.vault.configDir;
-		if (filePath === configDir || filePath.startsWith(configDir + '/')) return true;
-		return false;
+		// Excludes the plugin state folder and the Obsidian configuration directory.
+		// Delegates to the shared helper so the containment semantics live in one place.
+		return shouldExcludePath(
+			filePath,
+			normalizePath(this.plugin.settings.historyFolder),
+			this.plugin.app.vault.configDir
+		);
 	}
 
 	private passesPlatformGate(hook: Hook): boolean {
--- a/src/services/image-generation.ts
+++ b/src/services/image-generation.ts
@@ -271,8 +271,10 @@ export class ImageGeneration {
 			throw new Error(`Output path escapes the vault: "${outputPath}"`);
 		}
 
-		// Reject paths inside .obsidian/
-		if (normalized.split('/').includes('.obsidian')) {
+		// Reject paths inside the Obsidian configuration directory (default
+		// `.obsidian`, but the user may have renamed it)
+		const configDir = this.plugin.app.vault.configDir;
+		if (normalized === configDir || normalized.startsWith(configDir + '/')) {
 			throw new Error(`Output path cannot be inside the Obsidian configuration folder: "${outputPath}"`);
 		}
 
--- a/src/services/image-generation.ts
+++ b/src/services/image-generation.ts
@@ -3,7 +3,7 @@ import { Notice, App, MarkdownView, Modal, Setting, TextAreaComponent, TFile, no
 import { BaseModelRequest, GeminiClient, ModelClientFactory } from '../api';
 import { GeminiPrompts } from '../prompts';
 import { getErrorMessage } from '../utils/error-utils';
-import { ensureFolderExists } from '../utils/file-utils';
+import { ensureFolderExists, isPathInFolder } from '../utils/file-utils';
 import { t } from '../i18n';
 
 export class ImageGeneration {
@@ -272,9 +272,9 @@ export class ImageGeneration {
 		}
 
 		// Reject paths inside the Obsidian configuration directory (default
-		// `.obsidian`, but the user may have renamed it)
-		const configDir = this.plugin.app.vault.configDir;
-		if (normalized === configDir || normalized.startsWith(configDir + '/')) {
+		// `.obsidian`, but the user may have renamed it). Root-anchored, matching
+		// deep-research's write-path validator.
+		if (isPathInFolder(normalized, this.plugin.app.vault.configDir)) {
 			throw new Error(`Output path cannot be inside the Obsidian configuration folder: "${outputPath}"`);
 		}
 
--- a/src/services/vault-analyzer.ts
+++ b/src/services/vault-analyzer.ts
@@ -4,6 +4,7 @@ import { ModelClientFactory } from '../api';
 import { AgentsMemoryData } from './agents-memory';
 import { VaultAnalysisModal } from '../ui/vault-analysis-modal';
 import { collectFilesFromFolder } from '../utils/folder-walk';
+import { isPathInFolder } from '../utils/file-utils';
 import { t } from '../i18n';
 
 /**
@@ -290,7 +291,9 @@ export class VaultAnalyzer {
 	private getSampleFileNames(files: TFile[], limit: number = 20): string[] {
 		// Get files from different parts of the vault for diversity
 		const skipPaths = [this.plugin.settings.historyFolder, this.plugin.app.vault.configDir];
-		const filteredFiles = files.filter((f) => !skipPaths.some((skip) => f.path.startsWith(skip)));
+		// Root-anchored containment so a sibling like `gemini-scribe-backup/` or a
+		// renamed `_obsidian-notes/` is not wrongly excluded by a bare prefix match.
+		const filteredFiles = files.filter((f) => !skipPaths.some((skip) => isPathInFolder(f.path, skip)));
 
 		// Sort by modification time to get recent files
 		const sortedFiles = filteredFiles.sort((a, b) => b.stat.mtime - a.stat.mtime).slice(0, limit);
--- a/src/services/vault-analyzer.ts
+++ b/src/services/vault-analyzer.ts
@@ -240,8 +240,8 @@ export class VaultAnalyzer {
 		const indent = '  '.repeat(depth);
 		let structure = '';
 
-		// Skip system folders
-		const skipFolders = ['.obsidian', this.plugin.settings.historyFolder];
+		// Skip system folders (the Obsidian config dir may be renamed from `.obsidian`)
+		const skipFolders = [this.plugin.app.vault.configDir, this.plugin.settings.historyFolder];
 		if (skipFolders.includes(folder.path)) {
 			return '';
 		}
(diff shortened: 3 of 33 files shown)


src/services/vault-analyzer.ts
```diff
@@ -289,7 +289,7 @@ export class VaultAnalyzer {
 	 */
 	private getSampleFileNames(files: TFile[], limit: number = 20): string[] {
 		// Get files from different parts of the vault for diversity
-		const skipPaths = [this.plugin.settings.historyFolder, '.obsidian'];
+		const skipPaths = [this.plugin.settings.historyFolder, this.plugin.app.vault.configDir];
```
allenhutchison, line 293: **Bug (over-match): bare `startsWith` on the line below this one.** `getSampleFileNames` filters with `f.path.startsWith(skip)` — no `/` separator — so with `configDir = '_obsidian'` a user folder `_obsidian-notes/` is wrongly excluded, and `historyFolder = 'gemini-scribe'` wrongly excludes `gemini-scribe-backup/`. This is exactly the over-match class this PR fixes in `shouldExcludePath`; the historyFolder half pre-dates the PR, but since this line is being touched it's worth fixing here:

```ts
const filteredFiles = files.filter((f) => !skipPaths.some((skip) => f.path === skip || f.path.startsWith(skip + '/')));
```

Impact is limited (sample file names feeding the vault-analysis prompt), but it's a one-line fix.

src/services/image-generation.ts
```diff
@@ -271,8 +271,10 @@ export class ImageGeneration {
 			throw new Error(`Output path escapes the vault: "${outputPath}"`);
 		}
 
-		// Reject paths inside .obsidian/
-		if (normalized.split('/').includes('.obsidian')) {
+		// Reject paths inside the Obsidian configuration directory (default
+		// `.obsidian`, but the user may have renamed it)
+		const configDir = this.plugin.app.vault.configDir;
+		if (normalized === configDir || normalized.startsWith(configDir + '/')) {
```
allenhutchison: **Inconsistent protection semantics vs `deep-research.ts` in this same PR.** This check was narrowed from segment-anywhere (`normalized.split('/').includes('.obsidian')`) to vault-root-only, while `deep-research.ts` `validateAndNormalizeFilePath` kept the segment-anywhere loop (`for (const segment of segments) if (segment === configDir)`). So the two write-path validators now disagree: with a nested vault (or any subfolder named after the config dir, e.g. `subvault/.obsidian/`), deep-research rejects the write but image generation allows it — previously both rejected it.

Root-only is defensible (the *real* config dir only exists at the vault root, and root-only avoids over-matching folders merely named `.obsidian`), but the two services should pick one semantic. If root-only is the intent, deep-research's segment loop should be simplified to the same `=== configDir || startsWith(configDir + '/')` check; if nested-vault protection is the intent, this check should keep the segment scan against `configDir`.

src/services/hook-manager.ts
```diff
@@ -794,7 +794,8 @@ export class HookManager {
 		const stateFolder = normalizePath(this.plugin.settings.historyFolder) + '/';
 		if (filePath === stateFolder.slice(0, -1)) return true;
 		if (filePath.startsWith(stateFolder)) return true;
-		if (filePath.startsWith('.obsidian/')) return true;
+		const configDir = this.plugin.app.vault.configDir;
+		if (filePath === configDir || filePath.startsWith(configDir + '/')) return true;
```
allenhutchison: **Duplication: this exact pattern is now hand-rolled in four files.** `p === configDir || p.startsWith(configDir + '/')` appears here, in `image-generation.ts`, `obsidian-file-adapter.ts`, and `read-file-tool.ts`, while `shouldExcludePath(path, excludeFolder, configDir)` in `src/utils/file-utils.ts` already encapsulates it (this `isExcludedPath` is in fact semantically identical to `shouldExcludePath(filePath, this.plugin.settings.historyFolder, configDir)`). Consider either calling the existing helper or extracting a tiny `isInConfigDir(path, configDir)` next to it, so the next semantic tweak (e.g. the root-only vs segment-anywhere question raised on image-generation.ts) lands in one place instead of four.

Related nit: `read-file-tool.ts` computes `isObsidianPath` from `configDir` and then calls `shouldExcludePath` via the plugin wrapper, which re-checks the config dir — harmless, but the double check is another symptom of the missing shared helper.

src/utils/file-utils.ts
```diff
@@ -12,19 +12,35 @@ import type { Logger } from './logger';
...
 import { t } from '../i18n';
 
+/**
+ * Obsidian's default configuration directory name. Vaults can rename their
+ * config directory, so callers should prefer the vault's actual `configDir`
+ * (threaded through `shouldExcludePathForPlugin`); this is only the fallback
+ * used when no `configDir` is supplied.
+ */
+// eslint-disable-next-line obsidianmd/hardcoded-config-path -- documented fallback only; real callers pass vault.configDir via shouldExcludePathForPlugin
+export const DEFAULT_CONFIG_DIR = '.obsidian';
```
allenhutchison: **Latent gap: `createFileFilter` (line 75) still calls `shouldExcludePath` without `configDir`, so it silently rides on this hardcoded fallback.** Nothing in production currently calls `createFileFilter`, so this isn't reachable today — but it's exported, tested, and carries exactly the bug this PR fixes (renamed config dir not excluded, literal `.obsidian` folder over-matched) for any future caller. Consider threading a `configDir` parameter through `createFileFilter` too, which would let you drop `DEFAULT_CONFIG_DIR` and its eslint-disable entirely by making the third parameter required.

allenhutchison (commented):
Reviewed this with a multi-angle pass (line-by-line, removed-behavior audit, cross-file caller trace, reuse/altitude, conventions) and verified each candidate against the branch before posting. The core change is sound: `shouldExcludePathForPlugin` threads `vault.configDir` through, so all vault tools are fixed in one place, the ESLint rule re-enable is scoped correctly, and the new tests cover both the renamed-dir and over-match directions. Five findings survived verification — four are anchored inline; the fifth is a docs gap:

**Docs still describe the protected folder as literally `.obsidian`.** This PR updates `docs/guide/semantic-search.md` to say "`.obsidian` by default, or a renamed one", but five other user-facing docs describing the same protection were left stating the literal name: `docs/guide/agent-mode.md:796`, `docs/guide/deep-research.md:98` and `:159`, `docs/guide/getting-started.md:114`, `docs/guide/lifecycle-hooks.md:170`, and `docs/reference/settings.md:506`. Per the repo's documentation-maintenance mandate these should get the same one-line clarification. (The `docs/contributing/testing.md` mentions are install paths and fine as-is.)

src/services/vault-analyzer.ts
allenhutchison: Fixed in c5c0188. `getSampleFileNames` now routes the skip check through the shared `isPathInFolder` helper (`path === skip || path.startsWith(skip + '/')`), so `gemini-scribe-backup/` and a renamed `_obsidian-notes/` are no longer over-matched. Added as its own commit since it's a behavioral fix distinct from the refactor.

---

src/services/image-generation.ts
allenhutchison: Fixed in d3a2090 — went with **root-only** as you noted is defensible (the real config dir only exists at the vault root, and root-only avoids over-matching a content folder merely named `.obsidian`). `deep-research.ts`'s segment-anywhere loop is now replaced with the same root-anchored `isPathInFolder(normalizedPath, configDir)` check, so both write-path validators agree.

---

src/services/hook-manager.ts
allenhutchison: Fixed in d3a2090. Extracted `isPathInFolder(path, folder)` in `file-utils.ts` as the single root-anchored containment check. `hook-manager`'s `isExcludedPath` now delegates entirely to `shouldExcludePath(filePath, historyFolder, configDir)` (the semantic identity you pointed out), and `image-generation` / `obsidian-file-adapter` / `read-file-tool` call `isPathInFolder` — so the pattern lives in one place. Re the `read-file-tool` nit: `isObsidianPath` is still computed (it forces exclusion even for agent-session paths that fall under the config dir), but now via the shared helper; the subsequent `shouldExcludePath` re-check is inherent to that gate rather than a duplicated string check.

---

src/utils/file-utils.ts
allenhutchison: Fixed in d3a2090. Took the stronger option you suggested: `configDir` is now a **required** parameter on `shouldExcludePath`, and `createFileFilter` threads it through too — so `DEFAULT_CONFIG_DIR` and its eslint-disable are gone entirely. A missing configDir is now a compile error instead of a silent hardcoded-`.obsidian` fallback, which closes the latent gap for any future `createFileFilter` caller.

---

allenhutchison:
Thanks for the thorough pass — all five findings are addressed across three focused commits:

| # | Finding | Resolution | Commit |
|---|---------|-----------|--------|
| 1 | Docs still say literal `.obsidian` in 5 more files | Same one-line clarification applied to agent-mode, deep-research (×2), getting-started, lifecycle-hooks, settings reference | `78b9741` |
| 2 | `vault-analyzer` `getSampleFileNames` bare-`startsWith` over-match | Routed through `isPathInFolder` (root-anchored) | `c5c0188` |
| 3 | `p === configDir \|\| startsWith(configDir + '/')` hand-rolled in 4 files | Extracted `isPathInFolder`; hook-manager delegates to `shouldExcludePath`, the other three call the helper | `d3a2090` |
| 4 | deep-research (segment-anywhere) vs image-generation (root-only) disagreed | Picked **root-only** for both write-path validators via the shared helper | `d3a2090` |
| 5 | `createFileFilter` rides the hardcoded fallback | Made `configDir` **required**; threaded through `createFileFilter`; dropped `DEFAULT_CONFIG_DIR` + its eslint-disable | `d3a2090` |

One deviation from your suggestions worth flagging: the shared helper is named `isPathInFolder(path, folder)` rather than `isInConfigDir` — generic containment lets the same helper cover both the config dir and the plugin-state / exclude-folder checks (which is what fixed finding 2's over-match too).

Local pre-flight all green on the new tip: `lint` (0 errors), 3228 tests, `build`, `format-check`, `typecheck:test`.

---

PR #1099, opened 2026-07-03
refactor(rag): extract shared openRagStatusModal helper (#1067)

### LINKED ISSUE #1067: dry-violation: RagStatusModal construction duplicated between main.ts and rag-status-bar.ts
## What

Two places in the codebase construct `RagStatusModal` with four near-identical callbacks (open-settings, reindex, sync-pending, plus the modal opener), each ~30 lines:

- `src/main.ts:757–803` — inside the `gemini-scribe-rag-status` command callback.
- `src/services/rag-status-bar.ts:54–101` — inside the status-bar click handler.

The two blocks differ only in surface details:

- **Indexing handle**: `this.ragIndexing!` (`RagIndexingService` directly) vs `this.provider` (the `RagStatusProvider` interface the status-bar service was given). Both ultimately call `getDetailedStatus()`, `indexVault()`, and `syncPendingChanges()`.
- **App handle**: `this.app` vs `this.plugin.app`.
- **Translation keys**: `notice.main.ragIndexComplete` / `…ragIndexFailed` / `…ragSyncingPending` vs `notice.rag.indexingComplete` / `…indexingFailed` / `…syncingPending`. The English copy is essentially the same notice text in both branches — the duplication has spawned a parallel set of translation keys that say the same thing.
- The status-bar version has an extra "if indexing, open progress modal directly" preamble that the command-callback version doesn't have. That branch is real behavior, not duplication, but the rest of the callback structure is identical.

Refactoring this is borderline-mechanical but routed as an **issue, not a PR**, because the translation-key consolidation is a design decision (pick one set, deprecate the other, run `npm run translate` to regenerate the locale files) and the extraction needs to settle on a clean signature that both callers can use.

## Evidence

- `src/main.ts:757–803` — RAG-status command callback constructs the modal.
- `src/services/rag-status-bar.ts:54–101` — status-bar click handler constructs the same modal with parallel callbacks.
- Translation keys split across `notice.main.rag*` (used by `main.ts`) and `notice.rag.*` (used by `rag-status-bar.ts`) in `src/i18n/en.ts`.

## Proposed unit of work

1. Add a helper (e.g. `openRagStatusModal(app, indexing, pluginId)`) — either as a module function in `src/services/rag-status-bar.ts` next to the existing types, or as a new `src/services/rag-status-modal-launcher.ts` if you'd rather keep status-bar concerns separate. It owns the modal construction and all four callbacks.
2. Pick one of the two translation-key sets (suggest `notice.rag.*` since it's the more recently styled one) and update both call sites to share it; remove the other set from `src/i18n/en.ts`; run `npm run translate` to regenerate the locale files (or let the weekly i18n workflow do it after merge).
3. Replace both call sites with `openRagStatusModal(this.app, this.ragIndexing, this.manifest.id)` (and the status-bar's equivalent). The status-bar's "if indexing, open progress modal directly" branch stays at the call site — only the `else` branch becomes the shared helper call.
4. Cover the helper with a unit test that mounts the modal and asserts the reindex/sync callbacks are wired to the indexing handle.

## Out of scope

- Changing what `RagStatusModal` itself does. This is a duplication issue, not a modal redesign.
- The "if indexing, show progress modal" branch in the status-bar handler — that's status-bar-specific behavior, not duplicated logic.
- Touching `RagProgressModal` construction. The shared progress-modal callback inside the reindex flow can move into the helper, but `RagProgressModal` itself is fine.
- Renaming `RagIndexingService` / `RagStatusProvider`. The helper should take whichever interface both callers can satisfy (the smaller `RagStatusProvider` surface is probably right) without renaming the underlying classes.

---

allenhutchison:
## Summary

The RAG status modal was constructed in two places — the **RAG status** command in `main.ts` and the status-bar click handler in `rag-status-bar.ts` — with fully duplicated open-settings / reindex / sync-now callbacks. This extracts a single `openRagStatusModal(app, provider, pluginId)` helper and calls it from both sites, so the modal is wired up in exactly one place. Pure de-duplication with no behavior change.

Produced by the auto-dev pipeline from the approved plan on #1067.

Fixes #1067

## Changes

- `src/services/rag-status-bar.ts` — add exported `async function openRagStatusModal(app: App, provider: RagStatusProvider, pluginId: string): Promise<void>` that lazily imports `RagStatusModal` + `RagProgressModal`, constructs the modal with all four callbacks (open-settings, reindex, sync-now), and opens it. The status-bar click handler's `else` branch now delegates to it. The `if (status === 'indexing')` branch that opens `RagProgressModal` directly stays at the call site — that's status-bar-specific behavior, not shared logic.
- `src/main.ts` — the RAG-status command callback replaces its inline modal construction with `await openRagStatusModal(this.app, this.ragIndexing, this.manifest.id)`. The Ollama-unavailable and `ragIndexing` null guards stay at the call site. Drops the now-unused `getErrorMessage` and `openPluginSettingsTab` imports.
- `src/i18n/en.ts` — remove `notice.main.ragIndexComplete`, `notice.main.ragIndexFailed`, and `notice.main.ragSyncingPending`. Their messages were identical to the `notice.rag.{indexingComplete,indexingFailed,syncingPending}` keys the status-bar path (and now the command) already uses.
- `src/i18n/*.ts` (all 20 translated locales) + `src/i18n/translation-state.json` — drop the three removed keys so the locale files still typecheck against the narrowed `TranslationKey` union.
- `test/services/rag-status-bar.test.ts` (new) — covers `openRagStatusModal`: the status modal is constructed with the provider's detailed status and opened; the open-settings callback opens the plugin settings tab; the reindex callback opens the progress modal and triggers `indexVault()`; the sync callback calls `syncPendingChanges()` and only shows a notice when work was actually queued.

**Deviation from the plan:** the plan described the open-settings callback using an inline `@ts-expect-error app.setting.open()` pattern and mentioned a pending `openPluginSettingsTab` cleanup (PR #1066). That cleanup has since landed — both call sites already use `openPluginSettingsTab(app, pluginId)` — so the helper uses it directly rather than re-introducing the raw `@ts-expect-error` calls.

## Screenshots / Screencast

N/A — internal de-duplication refactor with no visual or behavioral change. The modal, its callbacks, and the notices shown are identical to before.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — approved plan on #1067
- [x] All CI checks pass (`npm test` — 3225 passing, `npm run build`, `npm run format-check`) — plus `npm run lint` (0 errors) and `npm run typecheck:test`
- [x] I have tested this change on Desktop — full unit suite passes; 5 new tests cover the extracted helper
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — pure code motion, no platform-specific APIs
- [x] Documentation has been updated (if applicable) — N/A; internal refactor with no user-facing surface

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01EBLhm7auRhimhtNuYWZGUs)_

## Summary by CodeRabbit

* **New Features**
  * Added a centralized entry point for the RAG status dialog, supporting settings, reindex, and sync actions from the app.
  * Updated localization to show state-specific RAG pause/resume messages consistently across multiple languages.
* **Bug Fixes**
  * Removed outdated RAG indexing/syncing “complete/failed/pending” notices from translations so the UI uses the newer wording.
  * Improved reliability by showing a user-facing message if opening the RAG status UI fails.
* **Tests**
  * Added automated coverage for RAG status dialog actions and resulting notifications.


allenhutchison:
Thanks @coderabbitai — both nitpicks addressed in `cd34065`:

1. **`src/main.ts` — missing error handling.** Wrapped the `gemini-scribe-rag-status` command's `openRagStatusModal` call in try/catch, mirroring the status-bar click handler and the sibling project commands: failures now log via `this.logger.error` and surface `notice.rag.uiError` (using `getErrorMessage`, matching the status-bar path's identical notice) instead of rejecting silently.
2. **`test/services/rag-status-bar.test.ts` — reindex Notice coverage.** Added two tests exercising the paths the shared helper centralizes: the progress-complete callback (`notice.rag.indexingComplete`) and the `indexVault()` rejection (`notice.rag.indexingFailed`).

Full pre-flight green (`build`, `npm test` — 3227 passing, `format-check`, `lint` 0 errors, `typecheck:test`).

The "several locales add `ragPaused`/`ragNotPaused`/`ragResumed` replacements" line in the walkthrough is a diff-context artifact — those keys already existed and were unchanged; this PR only *removes* the three redundant `notice.main.rag*` keys.

---

allenhutchison:
On the **Out of Scope Changes** pre-merge warning (the `main.ts` try/catch + `notice.rag.uiError`): flagging that this change was added in direct response to CodeRabbit's own review nitpick earlier in this PR ("Missing error handling around `openRagStatusModal`"), so it's the same reviewer requesting and then flagging it.

I'm treating it as **in scope** and keeping it: #1067 is about routing both entry points (the `gemini-scribe-rag-status` command and the status-bar click handler) through one shared `openRagStatusModal` helper *so they behave identically*. The status-bar path already wrapped its call in try/catch → `logger.error` + `notice.rag.uiError`; the command path didn't. Adding the matching guard brings the two consolidated call sites to parity, which is the core intent of the de-duplication rather than a separate concern. It's a 4-line guard with no new behavior beyond the error surface the sibling path already had.

@allenhutchison — if you'd rather keep this PR strictly to the mechanical extraction, splitting the command's error-handling into a one-line follow-up is trivial; just say the word and I'll move it. Otherwise it's ready as-is.

---

Issue #901, opened 2026-05-27
dry-violation: gemini/ollama clients duplicate the system-instruction assembly block


allenhutchison:
## What

`GeminiClient.buildGenerateContentParams()` and `OllamaClient.buildSystemInstruction()` independently run the same six-step recipe to assemble the system instruction passed to the model: load `AGENTS.md` memory off `plugin.agentsMemory`, swallow its errors via `plugin.logger.warn`, load skill summaries off `plugin.skillManager`, swallow those errors the same way, filter the skill list by `request.projectSkills` if set, and call `this.prompts.getSystemPromptWithCustom(...)` with the same six positional arguments in the same order.

The blocks are byte-identical apart from the variable name (`extReq` on the Gemini side, `request` on the Ollama side, because Gemini's branch lives inside a `'userMessage' in request` discriminator). This is the layer below #703's "provider abstraction" epic and below the per-provider package split that landed in #797 — that PR established the directory shape but did not consolidate the cross-provider duplication that the shape now makes visible.

It matters because the system-instruction layering is the plugin's primary contract with the model: every new layer (a new project field, a new memory source, a new skill scoping rule) currently has to be added in both files in lockstep. The clients also already share a `this.prompts: GeminiPrompts` instance, so the natural home for the helper is right there — there's no missing abstraction to invent, just one to extract.

## Evidence

Line numbers as of master (`a504409`).

**Gemini — inline in `buildGenerateContentParams`:**

- `src/api/providers/gemini/client.ts:212–246` — block runs when `isExtended` is true (i.e. `'userMessage' in request`).

```ts
// Load AGENTS.md memory if available
let agentsMemory: string | null = null;
if (this.plugin?.agentsMemory) {
    try {
        agentsMemory = await this.plugin.agentsMemory.read();
    } catch (error) {
        this.plugin.logger.warn('Failed to load AGENTS.md:', error);
    }
}

// Load available skill summaries for system prompt injection
let availableSkills: { name: string; description: string }[] = [];
if (this.plugin?.skillManager) {
    try {
        availableSkills = await this.plugin.skillManager.getSkillSummaries();
    } catch (error) {
        this.plugin.logger.warn('Failed to load skill summaries:', error);
    }
}

// Filter skills to project scope if active
if (extReq.projectSkills && extReq.projectSkills.length > 0) {
    availableSkills = availableSkills.filter((s) => extReq.projectSkills!.includes(s.name));
}

[...]

PR #1101, opened 2026-07-03
refactor: dedupe conversation-history normalization in GeminiClient

allenhutchison:
## Summary

Architecture-audit nightly sweep flagged: **DRY violation** in `src/api/providers/gemini/client.ts`.

`buildInteractionInput` (the Interactions transport path) and `buildContents` (the classic `generateContent` path) each iterated `conversationHistory` with the **identical** three-branch dispatch that normalizes each entry to a `Content`: canonical `{ role, parts }`, legacy `{ role, text }`, and legacy `{ role, message }`. This extracts a single private `normalizeHistoryEntry(entry): Content | null` helper and calls it from both loops — one pushes the returned `Content` directly, the other feeds it to `contentToSteps`. Pure behavior-preserving de-duplication, entirely within one file.

This is distinct from the already-resolved system-instruction-assembly duplication (#901); it's the history-entry normalization block.

## Changes

- `src/api/providers/gemini/client.ts` — add private `normalizeHistoryEntry(entry: Content): Content | null` (canonical/legacy-text/legacy-message dispatch, null for unrecognized entries); replace the duplicated inline dispatch in both `buildInteractionInput` and `buildContents` with a call to it.

## Screenshots / Screencast

N/A — internal de-duplication refactor with no visual or behavioral change; the same `Content` values are produced as before.

### Required

- [ ] This PR is linked to an approved issue where the approach was discussed with a maintainer — N/A; mechanical de-duplication from the automated architecture-audit sweep
- [x] All CI checks pass (`npm test` — 3220 passing, `npm run build`, `npm run format-check`) — plus `npm run typecheck:test`
- [x] I have tested this change on Desktop — full unit suite passes; the extracted helper produces identical output for all three entry shapes
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — pure data-normalization logic, no platform-specific APIs
- [x] Documentation has been updated (if applicable) — N/A; internal refactor with no user-facing surface

### AI-Generated Code

- [x] AI tool(s) used: Claude Code (architecture-audit sweep)

---
_Generated by [Claude Code](https://claude.ai/code/session_016iFzUYKHKva44wtUqEj4E9)_

### DIFF
--- a/src/api/providers/gemini/client.ts
+++ b/src/api/providers/gemini/client.ts
@@ -275,15 +275,26 @@ export class GeminiClient implements ModelApi {
 		if ('role' in entry && 'parts' in entry) {
 			return entry;
 		}
-		if ('role' in entry && 'text' in entry) {
-			const legacy = entry as Content & { text: string };
-			return { role: legacy.role === 'user' ? 'user' : 'model', parts: [{ text: legacy.text }] };
-		}
-		if ('role' in entry && 'message' in entry) {
-			const legacy = entry as Content & { role: string; message: string };
-			return { role: legacy.role === 'user' ? 'user' : 'model', parts: [{ text: legacy.message }] };
+		if ('role' in entry && ('text' in entry || 'message' in entry)) {
+			const legacy = entry as Content & { role?: string; text?: string; message?: string };
+			const text = legacy.text ?? legacy.message ?? '';
+			return { role: this.coerceHistoryRole(legacy.role), parts: [{ text }] };
 		}
 		return null;
+	}
+
+	/**
+	 * Map a legacy history role to a Gemini `Content` role. Only `user`/`model`
+	 * are valid Content roles; a `system` (or any other unexpected) role is
+	 * coerced to `model`, deliberately and with a warning rather than silently,
+	 * since replaying it as a model turn is a lossy fallback.
+	 */
+	private coerceHistoryRole(role: string | undefined): 'user' | 'model' {
+		if (role === 'user') return 'user';
+		if (role !== 'model') {
+			this.plugin?.logger.warn(`Unexpected conversation-history role "${role}", coercing to "model"`);
+		}
+		return 'model';
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


allenhutchison:
Two review requests before merge:

1. **Tighten the role coercion in `normalizeHistoryEntry`.** The legacy branches map any non-`'user'` role to `'model'`, so a `'system'` entry (which the `GeminiConversationEntry` type permits) would be silently replayed as a model turn. No current code path produces `'system'` entries, but now that this is the single normalization point, please handle it explicitly — either narrow the accepted role type or add an explicit branch/guard (a `logger.warn` on unexpected roles is fine) so the coercion is deliberate rather than incidental.

2. **Collapse the two legacy branches.** They differ only in which field they read; fold them into one branch (e.g. `const text = legacy.text ?? legacy.message` after checking `'text' in entry || 'message' in entry`), which also eliminates the inconsistent second cast that redundantly re-declares `role: string`. Same behavior, one less copy to keep in sync inside the helper that exists to end the duplication.

Everything else looks good — the extraction is behavior-preserving (branch order, strict `=== 'user'` mapping, reference identity, and silent-skip semantics all match the deleted code, and both duplicate sites are replaced).

allenhutchison:
Both addressed in 56e0acd:

1. **Role coercion made deliberate.** Extracted `coerceHistoryRole(role)`: `'user' → 'user'`, `'model' → 'model'`, and any other value (including the type-permitted `'system'`) coerces to `'model'` with a `logger.warn` rather than silently. Existing `'user'`/`'model'` output is unchanged; only unexpected roles now warn.

2. **Legacy branches collapsed.** Single branch guarded by `'text' in entry || 'message' in entry`, reading `legacy.text ?? legacy.message ?? ''`, which drops the redundant second cast that re-declared `role: string`.

`npm run format-check`, `npm run build`, `npm test` (3220 passing), and `npm run typecheck:test` all green locally.

---

PR #1206, opened 2026-07-13
refactor: consolidate legacy history-entry decode into a leaf helper (#1203)

### LINKED ISSUE #1203: dry-violation: legacy {text|message} history-entry decoding duplicated across 5 sites in 3 modules
## What

The backward-compat rule "a stored history entry may carry a top-level `text` **or** `message` field instead of canonical `parts`, and its text is `text ?? message`" is re-implemented from scratch in **five** places across **three** modules. Each site independently casts the entry to a `Content & { text?; message? }` ad-hoc shape and re-derives the text. This is a fragile, invariant-carrying duplication: if the legacy stored format ever changes (or a third legacy field appears), all five sites must be found and edited in lockstep, and there is nothing forcing them to stay consistent. `GeminiClient.normalizeHistoryEntry` — the one place that reads like the canonical decoder — is `private`, so nothing else can reuse it.

## Evidence

- `src/api/providers/gemini/client.ts:302-304` — `normalizeHistoryEntry`: `const legacy = entry as Content & { role?; text?; message? }; const text = legacy.text ?? legacy.message ?? '';` (private, unreusable)
- `src/api/providers/ollama/client.ts:376` — `const text = record.text ?? record.message;`
- `src/services/context-manager.ts:271-276` — `sanitizeContentsForTokenCount`: branches on `legacy.text` / `legacy.message`
- `src/services/context-manager.ts:528-531` — `compactHistory` split-boundary scan: `entry.parts?.[0]?.text || legacy.message || legacy.text`
- `src/services/context-manager.ts:601-605` — `summarizeConversation`: branches on `legacy.text` / `legacy.message`

## Proposed unit of work

- Add a leaf helper module (e.g. `src/utils/history-normalize.ts`) that owns the legacy shape in one place. Two small exports cover every call site:
  - `getLegacyEntryText(entry): string | undefined` — returns `text ?? message` (or the branch equivalent) for an entry that has no `parts`.
  - `normalizeToContent(entry): Content | null` — the `normalizeHistoryEntry` behavior (parts pass-through, legacy → `{ role, parts:[{text}] }`, unrecognized → null), so `GeminiClient` can delegate to it instead of keeping a private copy.
- Route all five sites through the helper. The Ollama client returns a different message shape (`{ role, content }`), so it should consume the shared **text extractor** (`getLegacyEntryText`) rather than the `Content`-shaped normalizer — keep the provider-specific assembly in each client.
- Delete the scattered `Content & { text?; message? }` casts in favor of the helper's typed signature.
- Add a small unit test covering the three input shapes (`parts`, legacy `text`, legacy `message`) plus the unrecognized-entry case, and a fixture for the `text`-and-`message`-both-present precedence (`text` wins).

## Out of scope

- Changing the on-disk / stored history format itself, or migrating old entries — this is purely about consolidating the read-side decode.
- The Ollama client's `{ role, content }` output shape and its role-coercion (`model`/`assistant` → `assistant`) — leave provider-specific assembly where it is; only the text-extraction step is shared.
- `context-manager`'s compaction/summarization logic beyond the one-line legacy-field reads.

---

allenhutchison:
## Summary

The backward-compat rule "a stored history entry may carry a top-level `text` **or** `message` field instead of canonical `parts`" was re-implemented from scratch at five sites across three modules (the Gemini and Ollama API clients, and the context manager), each with its own ad-hoc `Content & { text?; message? }` cast. If the legacy stored format ever changes, all five sites must be found and edited in lockstep with nothing forcing them to agree.

This is a pure read-side consolidation into a single leaf module — no stored-format or on-disk change.

Fixes #1203

## Changes

- **New `src/utils/history-normalize.ts`** (leaf module, never imports back into `api/` or `services/`) owning the legacy shape and decode precedence:
  - `getLegacyEntryText(entry)` — **nullish** precedence (`text ?? message`) for the API-replay paths, which treat an empty-string `text` as intentional.
  - `getLegacyEntryTextTruthy(entry)` — **truthiness** precedence (`text || message`) for the context-manager sites, which fall back to `message` when `text === ''`.
  - `normalizeToContent(entry, coerceRole)` — the former private `normalizeHistoryEntry` behavior (parts pass-through, legacy → `{ role, parts:[{text}] }`, unrecognized → `null`), with provider-specific role coercion passed in by the caller.
- `src/api/providers/gemini/client.ts` — `normalizeHistoryEntry` now delegates to `normalizeToContent`, keeping `coerceHistoryRole` local; ad-hoc cast removed.
- `src/api/providers/ollama/client.ts` — consumes `getLegacyEntryText`; the provider-specific `{ role, content }` assembly and `model`/`assistant` role-coercion stay in place.
- `src/services/context-manager.ts` — three sites (`sanitizeContentsForTokenCount`, `compactHistory` split-boundary scan, `summarizeConversation`) route through `getLegacyEntryTextTruthy`, preserving each site's exact truthiness precedence.
- `test/utils/history-normalize.test.ts` — new unit test covering the `parts`/`text`/`message` shapes, the unrecognized-entry → `null` case, and the empty-string edge that pins the nullish-vs-truthiness distinction.

### Precedence note (deviation from the plan's literal suggestion)

The approved plan suggested truthiness sites keep their behavior via a `getLegacyEntryText(entry) || ''`-style guard. That does **not** preserve the original behavior at the empty-string edge: when `text === ''` and `message` is set, the historical context-manager sites fall back to `message`, but `'' ?? message` returns `''`. To preserve behavior exactly, this PR provides a dedicated truthiness extractor (`getLegacyEntryTextTruthy`) alongside the nullish one, and the two differ only when `text === ''`. This is a small, documented adaptation that keeps the field knowledge fully centralized while preserving each call site's observable behavior.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#1203, plan approved)
- [x] All CI checks pass (`npm test` — 3431 passing, `npm run build`, `npm run format-check`, `npm run typecheck:test`, `npm run lint` — 0 errors)
- [x] I have tested this change on Desktop — behavior-preserving refactor; verified via the existing Gemini/Ollama/context-manager suites plus the new precedence-edge test. No new runtime surface to exercise beyond those decode paths.
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — no platform-specific APIs; pure TypeScript decode logic
- [ ] Documentation has been updated (N/A — internal refactor with no user-facing behavior change)

### AI-Generated Code

---

Produced by the auto-dev pipeline from the approved plan on #1203.

---
_Generated by [Claude Code](https://claude.ai/code/session_016KRagpkaEx9Wza5fZZ1rTo)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with older conversation-history formats across Gemini and Ollama.
  * Ensured legacy messages are consistently recognized, including entries using either text or message fields.
  * Preserved empty-message content where appropriate and improved fallback handling.
  * Improved conversation compaction and summarization when processing older history entries.

* **Tests**
  * Added coverage for legacy-history conversion, role mapping, fallback behavior, and invalid entries.

### DIFF
--- a/src/api/providers/gemini/client.ts
+++ b/src/api/providers/gemini/client.ts
@@ -31,6 +31,7 @@ import { GeminiPrompts } from '../../../prompts';
 import type { ObsidianGemini } from '../../../types/plugin';
 import { getDefaultModelForRole } from '../../../models';
 import { decodeHtmlEntities } from '../../../utils/html-entities';
+import { normalizeToContent } from '../../../utils/history-normalize';
 import type { GeminiClientConfig } from './config';
 import { ModelUseCase } from '../../model-use-case';
 import {
@@ -294,15 +295,7 @@ export class GeminiClient implements ModelApi {
 	 * canonical `{ role, parts }`. Returns null for unrecognized entries.
 	 */
 	private normalizeHistoryEntry(entry: Content): Content | null {
-		if ('role' in entry && 'parts' in entry) {
-			return entry;
-		}
-		if ('role' in entry && ('text' in entry || 'message' in entry)) {
-			const legacy = entry as Content & { role?: string; text?: string; message?: string };
-			const text = legacy.text ?? legacy.message ?? '';
-			return { role: this.coerceHistoryRole(legacy.role), parts: [{ text }] };
-		}
-		return null;
+		return normalizeToContent(entry, (role) => this.coerceHistoryRole(role));
 	}
 
 	/**
--- a/src/api/providers/ollama/client.ts
+++ b/src/api/providers/ollama/client.ts
@@ -34,6 +34,7 @@ import {
 import { GeminiPrompts } from '../../../prompts';
 import type { ObsidianGemini } from '../../../types/plugin';
 import type { OllamaClientConfig } from './config';
+import { getLegacyEntryText } from '../../../utils/history-normalize';
 
 export class OllamaClient implements ModelApi {
 	private client: Ollama;
@@ -373,7 +374,7 @@ export class OllamaClient implements ModelApi {
 
 		// Internal shape: { role, text } or { role, message }
 		if ('role' in record) {
-			const text = record.text ?? record.message;
+			const text = getLegacyEntryText(record);
 			if (typeof text !== 'string' || !text.trim()) return null;
 			const role = record.role === 'model' || record.role === 'assistant' ? 'assistant' : 'user';
 			return [{ role, content: text }];
--- a/src/services/context-manager.ts
+++ b/src/services/context-manager.ts
@@ -16,6 +16,7 @@ import { ModelClientFactory, ModelUseCase } from '../api';
 import { executeWithRetry } from '../utils/retry';
 import { createGoogleGenAI } from '../api/providers/gemini/google-genai-factory';
 import { truncateOldToolResults } from '../agent/agent-loop-helpers';
+import { getLegacyEntryTextTruthy } from '../utils/history-normalize';
 
 import contextSummaryPromptContent from '../../prompts/contextSummaryPrompt.hbs';
 
@@ -268,12 +269,11 @@ export class ContextManager {
 		const result: Content[] = [];
 		for (const entry of contents) {
 			if (!entry.parts || !Array.isArray(entry.parts)) {
-				// Legacy stored entries may have top-level text/message fields.
-				const legacy = entry as Content & { text?: string; message?: string };
-				if (legacy.text) {
-					result.push({ role: entry.role || 'user', parts: [{ text: legacy.text }] });
-				} else if (legacy.message) {
-					result.push({ role: entry.role || 'user', parts: [{ text: legacy.message }] });
+				// Legacy stored entries may have top-level text/message fields
+				// (truthiness precedence: an empty-string `text` falls back to `message`).
+				const legacyText = getLegacyEntryTextTruthy(entry);
+				if (legacyText) {
+					result.push({ role: entry.role || 'user', parts: [{ text: legacyText }] });
 				}
 				continue;
 			}
@@ -526,8 +526,7 @@ export class ContextManager {
 		// History entries may use parts[].text (API format) or message (stored format)
 		while (splitIndex > 0 && splitIndex < totalTurns) {
 			const entry = conversationHistory[splitIndex];
-			const legacy = entry as Content & { message?: string; text?: string };
-			if (entry.role === 'user' && (entry.parts?.[0]?.text || legacy.message || legacy.text)) {
+			if (entry.role === 'user' && (entry.parts?.[0]?.text || getLegacyEntryTextTruthy(entry))) {
 				break;
 			}
 			splitIndex--;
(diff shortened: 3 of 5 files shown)


allenhutchison:
## Fallback review

No external review arrived within the review window (CodeRabbit's automatic reviews are disabled on this repo, so it posted only a "Review skipped" notice), so this is the pipeline's own fresh-eyes review of the diff against the approved plan on #1203. It is a **self-review** — treat it as a signal, not independent sign-off.

**Examined:** the full diff (5 files, +200/−23) against the approved plan and the repo's behavior-preservation intent. The core risk in this refactor is that consolidating the five legacy `{ text | message }` decode sites silently changes one site's precedence, so each site was checked case-by-case against its original:

- `src/utils/history-normalize.ts` (new leaf module) — imports only the `Content` type from `@google/genai`; no back-imports into `api/`/`services/`, preserving the acyclic-graph invariant (the `Lint & Format` / `lint:cycles` job is green).
- **Gemini** `normalizeHistoryEntry` → `normalizeToContent(entry, coerce)`: `getLegacyEntryText(entry) ?? ''` reproduces the original `text ?? message ?? ''` (nullish) exactly, including the `text === ''` case where message is not used.
- **Ollama** client: `getLegacyEntryText` = `text ?? message` — matches the original `record.text ?? record.message`, and the `typeof/!trim()` guard downstream is unchanged.
- **context-manager** (sanitize / boundary-scan / summarize): `getLegacyEntryTextTruthy` = `text || message` reproduces the original truthiness branches, including the `text === '' → fall back to message` edge. The boundary-scan condition is boolean-only, so the extractor's text-first order vs. the original's message-first order yields the same result.
- Tests (`test/utils/history-normalize.test.ts`) pin the distinguishing `text === ''` edge for *both* extractors, the non-string field case, and all `normalizeToContent` branches (pass-through / legacy / null / caller-supplied role coercion).

**Fixed in this review:**

- nothing — no defects found. The documented deviation from the plan (a dedicated `getLegacyEntryTextTruthy` instead of a `getLegacyEntryText(...) || ''` guard) is in fact more correct than the plan's literal suggestion: `'' ?? message` returns `''`, which would drop the historical empty-string fallback. The split is the right call and is documented in the module header.

**For your judgement:**

- This is a read-side-only consolidation with no stored-format change, so the risk surface is entirely "does each site still decode identically?" — which the case-by-case check and the edge-pinning tests cover. Nothing else flagged.

---

CodeRabbit can be re-run on this PR at any time with a `@coderabbitai review` comment.

---

allenhutchison:
Addressed the CodeRabbit review nitpick on `src/utils/history-normalize.ts` (`normalizeToContent`'s canonical-shape branch not validating that `parts` is an array).

Fixed in `4e750a7`: the canonical branch now requires `Array.isArray(entry.parts)` before passing the entry through, matching the sibling guard in `context-manager.ts`. This is strictly safer — a malformed `{ parts: null }` entry that also carries a legacy `text`/`message` field now falls through to the legacy branch and recovers the text, instead of the mapper (`content.parts ?? []`) silently dropping it. Added two tests: a malformed non-array-`parts` entry (rejected) and one carrying a legacy `text` field (recovered).

Pre-flight green after the fix: build, `typecheck:test`, 3433 tests, format-check, lint (0 errors).

(Replying at PR level because the inline thread anchor went outdated once the fix commit changed those lines.)

---

PR #1207, opened 2026-07-14
refactor(tools): extract shared isFileInAgentScope predicate (#1204)

### LINKED ISSUE #1204: dry-violation: agent-scope file predicate (system-folder + project-root) duplicated across the two search tools
## What

The two vault search tools filter their candidate file set with the same two-clause predicate, written out inline and independently in each: exclude system folders (`shouldExcludePath(file.path, plugin)`), then scope to the active project (`projectRoot && !file.path.startsWith(projectRoot + '/')`). The `projectRoot + '/'` boundary check is a subtle, easy-to-get-wrong invariant (without the trailing slash, `Foo` would match `Foobar/…`) that is duplicated verbatim. If the agent-scoping rule ever changes, both tools must be edited in lockstep, and nothing enforces that they agree.

## Evidence

- `src/tools/vault/search-files-tool.ts:78-82` — `.filter()`: `if (shouldExcludePath(file.path, plugin)) return false; if (projectRoot && !file.path.startsWith(projectRoot + '/')) return false;`
- `src/tools/vault/search-file-contents-tool.ts:113-120` — same two guards as `continue` statements inside the scan loop.
- (Related but **not** the same invariant: `src/tools/vault/list-files-tool.ts:80-88` scopes by a user-supplied `folderPath`, not the project root, so its boundary check is a different concept — see Out of scope.)

## Proposed unit of work

- Add `isFileInAgentScope(file: TFile, plugin, projectRoot: string | undefined): boolean` to `src/tools/vault/utils.ts` (the established home for shared vault-tool helpers, alongside `resolvePathToFile` and the new `guardExcludedPath`). It returns `false` when the file is excluded or outside the project root, `true` otherwise.
- Use it as the single `.filter()` predicate in `search-files-tool` and as the loop guard in `search-file-contents-tool` (invert to `if (!isFileInAgentScope(...)) continue;`).
- Add a focused unit test for the boundary case: a file at `Foobar/x.md` must **not** be considered in-scope for `projectRoot = 'Foo'`.

## Out of scope

- `list-files-tool` — its boundary is a user-passed `folderPath`, a different filter than project scoping; deciding whether it should also adopt a shared helper is a separate call. Leave it untouched here.
- The search tools' matching/ranking logic, limits, and result shaping — only the file-eligibility predicate is being consolidated.
- Changing what "agent scope" means (the exclusion set or the project-root semantics); this is a pure extraction that preserves current behavior.

---

allenhutchison:
## Summary

The two vault search tools (`find_files_by_name`, `find_files_by_content`) each filtered their candidate file set with the same two-clause agent-scope predicate, written out inline and independently: exclude system folders (`shouldExcludePath`), then — when a project is active — scope to it via the `projectRoot + '/'` boundary check. That trailing-slash boundary is a subtle, easy-to-get-wrong invariant (without it, `projectRoot = "Foo"` would spuriously match `Foobar/note.md`) that was duplicated verbatim, so any change to agent scoping had to be made in lockstep across both tools with nothing enforcing that they agree.

This extracts a single `isFileInAgentScope(file, plugin, projectRoot)` predicate into `src/tools/vault/utils.ts` (the established home for shared vault-tool helpers, alongside `resolvePathToFile` and `guardExcludedPath`) and routes both tools through it. Pure behavior-preserving code motion — the filter logic is byte-for-byte equivalent to the inline versions it replaces.

Fixes #1204

## Changes

- `src/tools/vault/utils.ts` — add `isFileInAgentScope(file: TFile, plugin, projectRoot: string | undefined): boolean`, returning `false` when the file is in a protected system folder or outside the active project root, `true` otherwise. Docstring calls out the load-bearing `projectRoot + '/'` boundary.
- `src/tools/vault/search-files-tool.ts` — use it as the single `.filter()` predicate; drop the now-unused `shouldExcludePath` import.
- `src/tools/vault/search-file-contents-tool.ts` — use it as an inverted loop guard (`if (!isFileInAgentScope(...)) continue;`), collapsing the two separate `continue` guards into one; drop the now-unused `shouldExcludePath` import.
- `test/tools/vault/utils.test.ts` — new `isFileInAgentScope` describe block covering system-folder exclusion, no-project-root inclusion, in-/out-of-project-root scoping, and the `Foo` vs `Foobar/x.md` prefix boundary case.

**Out of scope** (per the approved plan): `list-files-tool` (its boundary is a user-supplied `folderPath`, a different concept), the search tools' matching/ranking logic, and any change to what "agent scope" means — this is a pure extraction that preserves current behavior.

## Screenshots / Screencast

N/A — internal refactor with no UI or user-visible behavior change.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#1204, plan approved)
- [x] All CI checks pass (`npm test` — 3418 passing, `npm run build`, `npm run format-check`, `npm run typecheck:test`, `npm run lint` — 0 errors)
- [x] I have tested this change on Desktop — behavior-preserving refactor; the extracted predicate is byte-for-byte equivalent to the inline logic, verified via the unchanged `search-files-tool` / `search-file-contents-tool` suites plus 5 new `isFileInAgentScope` unit tests (including the `Foo`/`Foobar` boundary). As a pure code-motion refactor there is no divergent runtime surface to drive end-to-end beyond what those tests exercise.
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — no platform-specific APIs touched; pure TypeScript predicate logic.
- [ ] Documentation has been updated (N/A — internal refactor with no user-facing behavior or settings change)

### AI-Generated Code

Produced by the auto-dev pipeline from the approved plan on #1204.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01P1E1XFqcKM6HqyuJtvkksq

---
_Generated by [Claude Code](https://claude.ai/code/session_01P1E1XFqcKM6HqyuJtvkksq)_

### DIFF
--- a/src/tools/vault/search-file-contents-tool.ts
+++ b/src/tools/vault/search-file-contents-tool.ts
@@ -1,8 +1,8 @@
 import { Tool, ToolResult, ToolExecutionContext } from '../types';
 import { ToolCategory } from '../../types/agent';
 import { ToolClassification } from '../../types/tool-policy';
-import { shouldExcludePathForPlugin as shouldExcludePath } from '../../utils/file-utils';
 import { getRawErrorMessageOr } from '../../utils/error-utils';
+import { isFileInAgentScope } from './utils';
 
 /**
  * Search for text content within files
@@ -110,12 +110,8 @@ export class SearchFileContentsTool implements Tool {
 
 			// Search through each file
 			for (const file of allFiles) {
-				// Skip system folders
-				if (shouldExcludePath(file.path, plugin)) {
-					continue;
-				}
-				// Scope to project root when active
-				if (projectRoot && !file.path.startsWith(projectRoot + '/')) {
+				// Skip files outside agent scope (system folders, or outside the active project root)
+				if (!isFileInAgentScope(file, plugin, projectRoot)) {
 					continue;
 				}
 
--- a/src/tools/vault/search-files-tool.ts
+++ b/src/tools/vault/search-files-tool.ts
@@ -1,8 +1,8 @@
 import { Tool, ToolResult, ToolExecutionContext } from '../types';
 import { ToolCategory } from '../../types/agent';
 import { ToolClassification } from '../../types/tool-policy';
-import { shouldExcludePathForPlugin as shouldExcludePath } from '../../utils/file-utils';
 import { getRawErrorMessageOr } from '../../utils/error-utils';
+import { isFileInAgentScope } from './utils';
 
 /**
  * Search for files by name pattern
@@ -76,8 +76,7 @@ export class SearchFilesTool implements Tool {
 
 			const projectRoot = context.projectRootPath;
 			const scopedMatches = allFiles.filter((file) => {
-				if (shouldExcludePath(file.path, plugin)) return false;
-				if (projectRoot && !file.path.startsWith(projectRoot + '/')) return false;
+				if (!isFileInAgentScope(file, plugin, projectRoot)) return false;
 				return regex.test(file.name) || regex.test(file.path);
 			});
 
--- a/src/tools/vault/utils.ts
+++ b/src/tools/vault/utils.ts
@@ -15,6 +15,20 @@ export function guardExcludedPath(normalizedPath: string, plugin: ObsidianGemini
 	return shouldExcludePath(normalizedPath, plugin) ? { success: false, error } : null;
 }
 
+/**
+ * Agent-scope predicate shared by the read-only vault search tools. A file is in
+ * scope when it is not inside a protected system folder (the plugin state folder
+ * or `.obsidian`) and — when a project is active — it lives under `projectRoot`.
+ *
+ * The `projectRoot + '/'` boundary is load-bearing: without the trailing slash a
+ * `projectRoot` of `Foo` would spuriously match `Foobar/note.md`.
+ */
+export function isFileInAgentScope(file: TFile, plugin: ObsidianGemini, projectRoot: string | undefined): boolean {
+	if (shouldExcludePath(file.path, plugin)) return false;
+	if (projectRoot && !file.path.startsWith(projectRoot + '/')) return false;
+	return true;
+}
+
 /** Plain, serializable description of a vault file or folder. */
 export interface VaultFileEntry {
 	name: string;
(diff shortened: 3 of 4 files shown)


allenhutchison:
## Fallback review

No external review arrived within the review window (CodeRabbit's automatic reviews are disabled on this repo, so it posted only a "Review skipped" notice), so this is the pipeline's own fresh-eyes review of the diff against the approved plan on #1204. It is a **self-review** — treat it as a signal, not independent sign-off.

**Examined:** the full diff (4 files, +50/−11) against the approved plan on #1204 and the repo's behavior-preservation intent. The core risk in a code-motion refactor is that the extracted `isFileInAgentScope` silently diverges from the two inline predicates it replaces, or that a now-unused import gets left behind:

- `src/tools/vault/utils.ts` (`isFileInAgentScope`, lines 26–30) returns `false` **iff** `shouldExcludePath(file.path, plugin)` is true **or** (`projectRoot && !file.path.startsWith(projectRoot + '/')`) — identical to the original two guards. So `!isFileInAgentScope(...)` reproduces each call site's pair of `return false` / `continue` guards exactly, including the load-bearing trailing-slash boundary.
- `search-files-tool.ts` and `search-file-contents-tool.ts` now reference only `isFileInAgentScope`; the previously-used `shouldExcludePathForPlugin` import was dropped from both and has no remaining references in either file (verified). `utils.ts` already imports `TFile` and `shouldExcludePath`, and its other helpers still use the latter, so no import breaks.
- `list-files-tool.ts` is deliberately left untouched — its boundary is a user-supplied `folderPath`, a distinct concept from project scoping — matching the plan's Out-of-scope. Its own `shouldExcludePath` usage is unaffected.
- `test/tools/vault/utils.test.ts` pins the load-bearing `Foo` vs `Foobar/x.md` trailing-slash boundary, plus system-folder exclusion, no-project-root inclusion, and in-/out-of-root scoping.
- CI is green on head `86d55b1` (`CI`, `Knip`, `Lint & Format` including `lint:cycles`).

**Fixed in this review:**

- nothing — no defects found. This is a pure behavior-preserving extraction; the predicate is logically equivalent to the inline code, the removed imports were genuinely unused, and the boundary invariant is covered by a dedicated test.

**For your judgement:**

- No runtime-behavior change to weigh. The only judgment call is the scope decision to leave `list-files-tool` out of the shared helper — a deliberate call the approved plan already made, not a gap in this PR.

---

CodeRabbit can be re-run on this PR at any time with a `@coderabbitai review` comment.

---

PR #1267, opened 2026-07-28
refactor(services): extract shared headless agent-turn driver (#1261)

### LINKED ISSUE #1261: dry-violation: headless agent-turn setup duplicated between ScheduledTaskRunner and HookRunner
## What

`ScheduledTaskRunner.run()` and `HookRunner.runAgentTask()` are two independent implementations of the same thing: "drive one headless agent turn." Both were already refactored to delegate the tool loop to `AgentLoop` (per the invariant that headless callers must consume `AgentLoop`, not reimplement it) — but everything *around* that call is still written twice, near-verbatim, in two files.

The duplicated span is roughly 60–70 lines each and covers: the agent-services-initialised guard, `createAgentSession` with the feature's `toolPolicy`, the `session.modelConfig` model override, `ToolExecutionContext` construction, `ModelClientFactory.createChatModel`, `getAutoApprovedTools` (with the same "headless auto-approves, so exclude ASK_USER" rationale comment in both), the turn preamble + `userMessage` assembly, the whole `ExtendedModelRequest` literal, the pre/post-call cancellation checks, the `AgentLoop.run` options object, and the `cancelled` / `exhausted` / `finalText` handling — including the exhaustion error message, whose wording is identical in both apart from the `[ScheduledTaskRunner]` vs `[HookRunner]` prefix and `Task`/`Hook` noun.

This is the second time these two files have converged on the same seam: #997 already extracted `HeadlessConfirmationProvider` out of exactly this pair. The setup around it never followed.

The practical cost is drift: a change to headless turn semantics (a new `ExtendedModelRequest` field, a different tool-exposure rule, a cancellation-check placement) has to be made twice, and there is no test that would catch the two copies diverging.

## Evidence

- `src/services/scheduled-task-runner.ts:30–125` — `ScheduledTaskRunner.run()`
- `src/services/hook-runner.ts:51–147` — `HookRunner.runAgentTask()`

Line-level pairs:

| Concern | scheduled-task-runner.ts | hook-runner.ts |
| --- | --- | --- |
| services guard | 30–33 | 52–55 |
| `createAgentSession` + toolPolicy | 38–42 | 58–62 |
| `modelConfig` override | 45–48 | 64–67 |
| `ToolExecutionContext` + `createChatModel` | 50–55 | 69–74 |
| `getAutoApprovedTools` + rationale comment | 56–60 | 75–79 |
| preamble / `userMessage` / `model` | 62–65 | 81–84 |
| `ExtendedModelRequest` literal | 67–78 | 86–100 |
| cancel → initial call → cancel | 80–82 | 103–105 |
| `AgentLoop.run` options | 86–110 | 107–125 |
| `exhausted` error + `finalText` | 110–123 | 126–139 |

The only genuine differences are: the session label (`Scheduled: <slug>` vs `Hook: <slug>`), the log prefix/noun in error strings, `HookRunner` rendering its prompt through `renderPrompt(...)` first, and `HookRunner` passing `projectSkills` for `hook.enabledSkills`. Output handling then diverges legitimately (the task always writes output; the hook writes only when `outputPath` is set) — that part is *not* duplicated and should stay in the callers.

## Proposed unit of work

- Add a leaf module (e.g. `src/services/headless-agent-turn.ts`) exporting a single function — something like `runHeadlessAgentTurn(plugin, spec, isCancelled): Promise<string | undefined>` — that owns the whole duplicated span and returns `finalText` (or `undefined` when cancelled).
- Give it a small parameter object covering the four real differences: `sessionLabel`, `logPrefix` / `subjectNoun` (for the exhaustion error), `prompt` (already rendered by the caller), `toolPolicy`, `model`, `maxIterations`, and optional `projectSkills`. Keep the `ExtendedModelRequest` assembly and the `AgentLoop` options inside the helper.
- Rewrite `ScheduledTaskRunner.run()` to call it and then do only its own output write (`resolveOutputPath` + frontmatter header + `writeHeadlessOutput`).
- Rewrite `HookRunner.runAgentTask()` to call it and then do only its own conditional output write.
- Add `test/services/headless-agent-turn.test.ts` covering the shared path once: cancellation before/after the initial call, the `exhausted` error, the no-tool-calls fallback to `initialResponse.markdown`, and that `getAutoApprovedTools` (not the full registry) is what reaches `availableTools`. The existing `scheduled-task-manager` / hook tests then only need to assert their own output-writing behavior.
- Keep the `HeadlessConfirmationProvider` and `headless: true` wiring inside the helper so the "headless never surfaces ASK_USER tools" invariant has exactly one enforcement site.

## Out of scope

- The non-`agent-task` hook actions (`summarize`, `rewrite`, `command`) — they don't drive an agent turn and shouldn't be folded in.
- Output-path resolution and `writeHeadlessOutput`; those already share `headless-run-output.ts` and their per-caller differences are real.
- Any change to `AgentLoop` itself, to headless iteration defaults, or to the tool-exposure policy. This is code motion, not a semantics change.
- `src/services/headless-confirmation-provider.ts` currently has no test file; adding one is worth doing but belongs to the untested-modules issue, not here.

---

allenhutchison:
## Summary

`ScheduledTaskRunner.run()` and `HookRunner.runAgentTask()` were two near-verbatim implementations of the same thing — "drive one headless agent turn." Both already delegated the tool loop to `AgentLoop`, but the ~65 lines *around* that call were written twice: the agent-services guard, `createAgentSession` with the feature's `toolPolicy`, the `session.modelConfig` override, `ToolExecutionContext` construction, `ModelClientFactory.createChatModel`, `getAutoApprovedTools` (with the same rationale comment in both), the turn preamble, the whole `ExtendedModelRequest` literal, the pre/post-call cancellation checks, the `AgentLoop.run` options, and the `cancelled` / `exhausted` / `finalText` handling.

The duplication had already started to cost: the exhaustion error message was maintained in two places, and there was no test that would have caught the two copies diverging.

This extracts the whole span into a new leaf module, `src/services/headless-agent-turn.ts`, exporting a single `runHeadlessAgentTurn()`. The four real differences become a spec parameter object. Both callers keep only their own output-writing behavior — which genuinely differs and should stay split: the scheduled task always writes its result and treats empty output as a *failure*; a hook writes only when `outputPath` is set and treats empty output as a silent no-op.

This is the second extraction from this same file pair — #997 split out `HeadlessConfirmationProvider`, and the setup around it never followed. The `headless: true` + `HeadlessConfirmationProvider` wiring now lives inside the helper, so the "headless never surfaces ASK_USER tools" invariant has exactly one enforcement site.

Produced by the auto-dev pipeline from the plan approved on the issue.

Fixes #1261

## Changes

- **`src/services/headless-agent-turn.ts` (new, 167 lines)** — `runHeadlessAgentTurn(plugin, spec, isCancelled)`. Owns the entire duplicated span; returns `finalText`, or `undefined` when cancelled. Leaf module (imports the session/factory/loop layers; nothing imports the runners back), per the shared-leaf-helper rule in `coding.md`.
- **`HeadlessAgentTurnSpec`** — `sessionLabel`, `logPrefix`, `subjectNoun` + `subjectName` (so the exhaustion message has one source instead of two drifting copies), `prompt` (callers render templates first), `toolPolicy`, `model`, `maxIterations`, `projectSkills`.
- **`src/services/scheduled-task-runner.ts`** — 133 lines lighter; `run()` now builds the spec, calls the helper, and does only its `resolveOutputPath` + frontmatter header + `writeHeadlessOutput`.
- **`src/services/hook-runner.ts`** — 123 lines lighter; `runAgentTask()` now renders the prompt, builds the spec (passing `hook.enabledSkills` as `projectSkills`), calls the helper, and does only its conditional output write.
- **`test/services/headless-agent-turn.test.ts` (new, 24 tests)** — covers the shared path once.
- **`test/services/{scheduled-task-runner,hook-runner}.test.ts`** — trimmed the assertions that are now shared-path assertions, so they live in exactly one place. Each caller's own mapping and output-writing tests stay.

### Behavior-equivalence notes

Two places where the code motion could have changed behavior, both checked:

- **`projectSkills` on the scheduled-task request.** The scheduled path previously omitted the key entirely; it now passes `projectSkills: undefined`. Consumption is by truthiness (`src/prompts/gemini-prompts.ts:212` — `if (request.projectSkills && request.projectSkills.length > 0)`), never by key presence, so explicit-`undefined` and absent are equivalent.
- **Hook guard ordering.** `renderPrompt()` now runs before the agent-services guard (the guard moved inside the helper). `renderPrompt` is pure string substitution with no failure mode, and the thrown message is byte-identical (`[HookRunner] Agent services not initialised`).

## Screenshots / Screencast

N/A — no UI surface. Pure internal code motion.

## Testing

**Automated — full pre-flight green:**

- `npm test` — **164 files, 3534 tests passing**
- `npm run build`, `npm run typecheck:test`, `npm run format-check` — clean
- `npm run lint` — 0 errors (38 pre-existing warnings, none in touched files)
- `npm run lint:cycles` — **no circular dependency**, confirming the new module is a true leaf
- `npm run knip` — clean

**The strongest evidence this is behavior-preserving:** the pre-existing `scheduled-task-runner` and `hook-runner` suites — **56 tests** — passed **unmodified** against the refactored code, before any test was trimmed. The trim came afterward and only removed assertions that the new shared suite now makes.

**New coverage** (`test/services/headless-agent-turn.test.ts`, 24 tests): the services guard for each of the three handles; cancellation before the call, during the call, and reported by `AgentLoop`; the no-tool-calls fallback to `initialResponse.markdown` (including the null → `''` case, which is what lets callers distinguish "cancelled" from "ran but produced nothing"); the exhaustion message with its caller prefix, subject noun, cap and actual count; `maxIterations` forwarding; session/model/preamble/request construction; `projectSkills` include-list semantics; `HeadlessConfirmationProvider` wiring; and a direct guard that `getAutoApprovedTools` — **not** `getEnabledTools` — is what reaches `availableTools`.

**Behavioral verification not run in sandbox — needs a manual check.** Firing a real scheduled task or hook end-to-end needs a live Obsidian vault and a Gemini API key, neither of which this sandbox has. The change is pure code motion with no runnable-surface change, and the unmodified-suite result above is the substitute evidence, but no scheduled task or hook was actually fired.

**Mobile** — no platform-specific APIs added or moved; the code is provider/session plumbing only.

## Documentation

No documentation updates. This changes no user-facing behavior, setting, command, or public surface — it is internal code motion between two service modules. Flagging explicitly given the repo's mandatory-docs policy: this is the documented "N/A, and here's why" case, and the approved plan called it out the same way.

### Required

- [ ] I have tested this change on Desktop — **not done**: no live vault in the sandbox (see Testing)
- [x] Documentation has been updated (if applicable) — N/A, rationale above

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_015bZrjuRSZAniKh3UJntxRo)_

## Summary by CodeRabbit

* **New Features**
  * Added a shared headless agent execution driver used by scheduled tasks and automated hooks.
  * Supports per-run model overrides, tool policy, project skill constraints, iteration limits, and headless auto-confirmations.

* **Bug Fixes**
  * Canceled automated runs now stop cleanly without writing output.

* **Tests**
  * Added comprehensive coverage for the shared headless execution flow (cancellation, tool handling, model selection, and exhaustion errors).
  * Reduced/relocated overlapping runner test assertions to avoid duplication.

### DIFF
--- a/src/services/headless-agent-turn.ts
+++ b/src/services/headless-agent-turn.ts
@@ -1,0 +1,167 @@
+import { getActiveChatModel } from '../models';
+import type { ObsidianGemini } from '../types/plugin';
+import type { FeatureToolPolicy } from '../types/tool-policy';
+import { DestructiveAction } from '../types/agent';
+import { ToolExecutionContext } from '../tools/types';
+import { ModelClientFactory } from '../api';
+import { ExtendedModelRequest } from '../api/interfaces/model-api';
+import { formatLocalTimestamp } from '../utils/format-utils';
+import { buildTurnPreamble } from '../utils/turn-preamble';
+import { AgentLoop, DEFAULT_HEADLESS_MAX_ITERATIONS } from '../agent/agent-loop';
+import { HeadlessConfirmationProvider } from './headless-confirmation-provider';
+
+/**
+ * Shared driver for a single headless agent turn, used by both
+ * `ScheduledTaskRunner` and `HookRunner`.
+ *
+ * Both callers need exactly the same thing — create a temporary session under a
+ * feature tool policy, send one non-streaming request, hand any tool calls to
+ * `AgentLoop`, and come back with the final text — and the two copies had
+ * already drifted in wording. Keeping the whole span here means a change to
+ * headless turn semantics (a new `ExtendedModelRequest` field, a different
+ * tool-exposure rule, a cancellation-check placement) is made once.
+ *
+ * What deliberately stays with the callers: the output write. The scheduled task
+ * always writes its result and treats empty output as a failure; a hook writes
+ * only when `outputPath` is set and treats empty output as a silent no-op. Those
+ * differences are real, and `headless-run-output.ts` already shares the
+ * mechanical parts.
+ *
+ * This is a leaf module: it imports the session/factory/loop layers and nothing
+ * imports the runners back, per the shared-leaf-helper rule in `coding.md`.
+ */
+export interface HeadlessAgentTurnSpec {
+	/** Session title, e.g. `Scheduled: my-task` or `Hook: my-hook`. */
+	sessionLabel: string;
+	/** Log/error prefix identifying the caller, e.g. `[ScheduledTaskRunner]`. */
+	logPrefix: string;
+	/** Noun used in the exhaustion error, e.g. `Task` or `Hook`. */
+	subjectNoun: string;
+	/** Human-readable identifier of the unit of work, used in error messages. */
+	subjectName: string;
+	/** The prompt to send. Callers render any templates before handing it over. */
+	prompt: string;
+	/** Per-feature tool policy; when unset the session inherits the global policy. */
+	toolPolicy?: FeatureToolPolicy;
+	/** Model override; falls back to the active chat model. */
+	model?: string;
+	/** Tool-iteration cap; falls back to `DEFAULT_HEADLESS_MAX_ITERATIONS`. */
+	maxIterations?: number;
+	/**
+	 * Skill include-list passed through to the model API. An empty or absent
+	 * list means "no filter" — the run sees every available skill.
+	 */
+	projectSkills?: string[];
+}
+
+/**
+ * Drive one headless agent turn to completion.
+ *
+ * @returns the final response text, or `undefined` if the run was cancelled.
+ * @throws if agent services are not initialised, or if the tool-iteration
+ *         budget is exhausted without producing a response.
+ */
+export async function runHeadlessAgentTurn(
+	plugin: ObsidianGemini,
+	spec: HeadlessAgentTurnSpec,
+	isCancelled: () => boolean
+): Promise<string | undefined> {
+	if (!plugin.sessionManager || !plugin.toolRegistry || !plugin.toolExecutionEngine) {
+		throw new Error(`${spec.logPrefix} Agent services not initialised`);
+	}
+
+	// Create a headless session bound to the caller's tool policy. When there is
+	// no policy, the session inherits the global plugin policy — the registry
+	// filter is permission-driven, so a run without a policy sees the same tools
+	// as an interactive agent session.
+	const session = await plugin.sessionManager.createAgentSession(spec.sessionLabel, {
+		toolPolicy: spec.toolPolicy,
+		requireConfirmation: [] as DestructiveAction[],
+	});
+
+	// Propagate the per-run model override so follow-up requests also use the
+	// right model via session.modelConfig.
+	if (spec.model) {
+		session.modelConfig = { model: spec.model };
+	}
+
+	const toolContext: ToolExecutionContext = {
+		plugin,
+		session,
+		featureToolPolicy: spec.toolPolicy,
+	};
+	const modelApi = ModelClientFactory.createChatModel(plugin);
+	// Headless runs auto-approve confirmations, so only expose tools the user
+	// explicitly opted into (APPROVE under the layered policy). ASK_USER tools
+	// are excluded — exposing them would silently bypass the user's "ask first"
+	// intent because there's no UI to ask on. To allow an ASK_USER tool in a
+	// headless run, the run's toolPolicy must explicitly upgrade it (preset or
+	// per-tool override).
+	const availableTools = plugin.toolRegistry.getAutoApprovedTools(toolContext);
+
+	// Prepend a turn preamble so the model has accurate "now" awareness.
+	const startedAt = formatLocalTimestamp(session.created);
+	const userMessage = buildTurnPreamble(formatLocalTimestamp(new Date())) + spec.prompt;
+	const model = spec.model ?? getActiveChatModel(plugin.settings);
+
+	const initialRequest: ExtendedModelRequest = {
+		kind: 'extended',
+		userMessage,
+		conversationHistory: [],
+		model,
+		temperature: plugin.settings.temperature,
+		topP: plugin.settings.topP,
+		prompt: '',
+		availableTools,
+		renderContent: false,
+		sessionStartedAt: startedAt,
+		// The model API uses projectSkills as its include-list when filtering the
+		// registered skill set. Empty list ⇒ no filter, so the run sees every
+		// available skill (matches the documented "leave blank to inherit"
+		// semantics).
+		projectSkills: spec.projectSkills?.length ? spec.projectSkills : undefined,
+	};
+
+	if (isCancelled()) return undefined;
+	const initialResponse = await modelApi.generateModelResponse(initialRequest);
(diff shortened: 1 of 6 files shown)


allenhutchison:
Both nitpicks triaged — one fixed, one filed as a follow-up.

**1. Hard-coded `DEFAULT_HEADLESS_MAX_ITERATIONS: 20` in the mock — valid, fixed in a1d2bd9.**

Agreed, and it was a real staleness hazard: the mock supplied the constant the driver reads, so changing the real default would have left `expect(maxIterations: 20)` passing while the driver's behavior drifted. Took the suggested `importOriginal` approach — only `AgentLoop` is stubbed now, every other export comes from the real module — and the assertion compares against the imported constant.

I checked the fix isn't vacuous rather than assuming: substituting a deliberately wrong value into the assertion makes it fail (`expected … "maxIterations": 999999`), which confirms the constant resolves to a real number and is genuinely compared. Had `importOriginal` returned `undefined`, the assertion would have silently passed against `undefined` — worth ruling out. Test time is unchanged (~1.2s), so pulling in the real module costs nothing.

One adjustment beyond the suggestion: the exhaustion test also used `20`, but as an *explicit* `maxIterations` override rather than the default. Leaving it as `20` made it read as a default-cap test. It now uses a deliberately non-default cap (`cap 7, ran 8`), so it proves the message reports the configured value instead of coincidentally matching the shared default.

**2. `fellBack` / `loopAborted` returned like a real answer — correct, and filed as #1268 rather than fixed here.**

This one is accurate and worth acting on. Both flags document `markdown` as a notice the caller "may display but should not persist" (`src/agent/agent-loop.ts:161–198`), and the driver reads neither — so `ScheduledTaskRunner` writes the notice into a vault note with normal `scheduled_task:` / `ran_at:` frontmatter, indistinguishable from a real run. For a daily task, that notice silently becomes the day's report.

Keeping it out of this PR, for the reason you identified yourself: the behavior matches the pre-refactor copies exactly, so it's pre-existing, and #1261's approved plan scoped this PR to pure code motion with "any change to `AgentLoop` … or semantics" explicitly out of scope. Folding it in would also mean deciding what a scheduled task *should* do with a notice — skip the write and count a failure, write it with an `incomplete` marker, or log only — and that's a maintainer call, not a review-round fix. #1268 lays out those options.

Worth noting the consolidation this PR performs is what makes that fixable at all: the flags now go ignored in one place instead of two.

**Not addressed:** the `projectSkills`-on-`ScheduledTask` note, which you self-marked as likely invalid — correctly so; `ScheduledTask` has no skill include-list, and the scheduled path passes no `projectSkills` by design.

CI is green on the pre-fix commit (all 7 checks) and re-running on a1d2bd9.

---

Issue #1262, opened 2026-07-28
missing-tests: 20 source modules (~3.9k lines) have no test file, concentrated in src/ui/


allenhutchison:
## What

The test suite is broad (164 files, 3523 tests) but has a structural hole: **20 non-type, non-barrel, non-generated source modules — about 3,900 lines — are never imported by any test.** They are heavily concentrated in `src/ui/` (modals and agent-view helpers), and they include several modules that gate destructive or permission-sensitive behavior.

This is an umbrella issue because the findings share one root cause (Obsidian `Modal` / view classes are awkward to instantiate under jsdom, so they got skipped) and a reviewer should weigh them against each other rather than pick them up in isolation. It is a tracker: file one sub-issue per module (or per cluster) when picking it up.

Note the repo *does* have working patterns for this — `test/ui/management-modal-base.test.ts`, `test/ui/settings-general.test.ts`, and `test/ui/agent-view/*` all exercise Obsidian UI classes against the `__mocks__/obsidian` fixture. The gap is coverage, not missing infrastructure.

## Evidence

Detection: collect every `from '…src/…'` and `vi.mock('…src/…')` specifier across `test/`, resolve barrel imports (e.g. `src/tools/vault` covers `src/tools/vault/*`), then diff against `src/**/*.ts`. Excluded from the list below: `src/i18n/*` (AI-generated), `src/services/generated-help-references.ts` (build-generated), type-only modules (`src/types/*`, `src/services/skill-types.ts`), barrels (`src/index.ts`, `src/api/index.ts`, …), and modules covered transitively through a re-export (`src/services/hook-types.ts` — `renderPrompt` is tested via `hook-manager`).

**Priority 1 — behavior-gating, test them first**

| Module | Lines | Why it matters |
| --- | --- | --- |
| `src/ui/agent-view/agent-view-tools.ts` | 380 | The `AgentView` adapter over `AgentLoop` — tool batching, confirmation dispatch, and `inlineData` injection all land here. |
| `src/services/headless-confirmation-provider.ts` | 33 | Tiny, but it is the auto-approve provider for every headless run (hooks, scheduled tasks). A regression here silently approves things unattended. |
| `src/ui/components/tool-policy-editor.ts` | 243 | Edits the per-feature tool policy that decides what a hook or scheduled task may run. |
| `src/ui/yolo-confirmation-modal.ts` | 72 | The bypass-confirmations surface. |

**Priority 2 — large, actively-changing UI**

[...]

PR #1280, opened 2026-07-30
feat(agent): replace native confirm() on session delete with an inline row confirm

### LINKED ISSUE #1275: Obsidian-API drift: session delete uses a native `confirm()` dialog instead of an Obsidian modal
## What

`coding.md`'s "Obsidian API first" rule says to prefer built-in Obsidian API over low-level operations. Deleting an agent session from the session list is the one place in `src/` that reaches for the browser's native `confirm()` instead — a blocking dialog rendered by the host chrome, outside Obsidian's modal system and outside the plugin's theming. It is already flagged in place with an eslint suppression and a `TODO`, so the intent to replace it exists; it just hasn't been picked up.

Three things make it worth doing rather than leaving as a permanent `TODO`:

1. **It is destructive.** This is the confirmation gate on deleting a session's history file — the one dialog in the plugin where a mis-click is unrecoverable.
2. **Mobile.** `confirm()` behavior in the Obsidian mobile webview is not something the plugin controls, and every other confirmation surface here is a real Obsidian `Modal`. This is the outlier.
3. **It is the only one.** A grep for `confirm(` / `alert(` across `src/` returns exactly this site — so fixing it retires the `no-alert` suppression from the codebase entirely rather than fixing one of many.

Routed as an issue rather than a PR because there is no shared confirm-modal component to drop in, and the repo has **two** established patterns to choose between — that choice is the design decision.

## Evidence

`src/ui/agent-view/session-list-modal.ts:241–247`:

```ts
deleteBtn.addEventListener('click', (e) => {
    e.stopPropagation();
    // eslint-disable-next-line no-alert -- TODO: replace with Obsidian confirmation Modal
    if (confirm(t('agent.sessionList.deleteConfirm', { title: session.title }))) {
        void this.deleteSession(session);
    }
});
```

- Only occurrence in the tree: `grep -rn '\bconfirm(\|\balert(' src/ --include='*.ts'` returns this line and nothing else.
- It is also the sole `no-alert` entry in the codebase's 50 eslint suppressions.

Existing patterns it could adopt, both already in the repo:

- **Two-step inline confirm** — `src/ui/scheduler-management-modal.ts` and `src/ui/hook-management-modal.ts` turn the delete button into a confirm button in place. Styles already exist (`.gemini-scheduler-confirm-delete`, `styles.css:5127`). No new component, no new modal, and it works inside an already-open modal — which matters here, since the session list *is* a modal and stacking a second one is the awkward case.
- **Dedicated modal** — `src/ui/yolo-confirmation-modal.ts` is the closest thing to a reusable confirmation surface, though it is purpose-built rather than generic.

## Proposed unit of work

- Pick a pattern. The two-step inline confirm is the lower-risk one: it matches what the other two management modals already do, needs no new file, and avoids a modal-over-modal interaction. A generic `ConfirmModal` is the higher-value option only if a second caller is expected.
- Replace the `confirm()` call in `session-list-modal.ts` with the chosen pattern and drop the `eslint-disable-next-line no-alert` comment.
- Keep the existing i18n key `agent.sessionList.deleteConfirm` as the message; a two-step button needs one additional short label key in `src/i18n/en.ts` (the confirm state), added with its `context` per the i18n rule. Per-language files regenerate from the `Update UI translations` workflow — no manual translation needed.
- Style the confirm state from the `--gs-*` tokens; reuse `--gs-error` as the scheduler delete button does rather than introducing new color.
- If a generic `ConfirmModal` is chosen instead, put it in `src/ui/components/` alongside `management-modal-base.ts` and give it a test in the `test/ui/management-modal-base.test.ts` shape.

## Out of scope

- Migrating the scheduler and hook modals' existing confirm flows. They already comply; this issue is only about the one non-compliant site.
- Adding an undo path for session deletion, or changing what `deleteSession` actually does.
- The other 49 eslint suppressions in `src/`, which are unrelated rules with their own rationales.
- Any broader redesign of the session list.

---

allenhutchison:
## Summary

Deleting an agent session from the session list was gated by the browser's native `confirm()` — the only such call left in `src/`, carrying its own `no-alert` eslint suppression and a `TODO`. It renders outside Obsidian's modal system and its mobile-webview behavior isn't the plugin's to control, on the one dialog in the plugin where a mis-click is unrecoverable.

This replaces it with a **row-level inline confirm**: clicking the trash icon swaps that row's action buttons for the existing `Delete session "…"?` prompt plus **Delete** / **Cancel**; Cancel rebuilds the buttons. The list keeps its scroll position and filter state, and no second modal is stacked over the session list.

Two notes on the approach, since the issue's framing was slightly off:

- The issue described the existing pattern as modals that "turn the delete button into a confirm button in place." They don't — `ManagementModalBase.confirmDelete()` empties `contentEl` and renders a full-pane confirm screen. There was no row-level in-place confirm in the repo, so this is a new micro-pattern rather than reuse, and it's worth saying so plainly.
- `SessionListModal extends Modal` directly, not `ManagementModalBase`, so there was nothing to inherit either way. A full-pane swap here would discard the filtered list's scroll position to ask one question — the same re-render dance `deleteSession()` already performs after a successful delete.

This retires the last `no-alert` suppression from the codebase: `grep -rn '\bconfirm(\|\balert(' src/ --include='*.ts'` is now empty, and `npm run lint` passes with the suppression deleted rather than relocated.

Built by the auto-dev pipeline from the plan approved on the issue.

Fixes #1275

## Changes

- `src/ui/agent-view/session-list-modal.ts` — replace the `confirm()` branch with the inline confirm; delete the `eslint-disable-next-line no-alert` and its `TODO`. Row-action rendering moves into `renderSessionActions()` so cancelling rebuilds buttons with live handlers instead of un-hiding stale ones. `showDeleteConfirm()` renders the prompt and the two decision buttons; `dismissDeleteConfirm()` restores. State is one `pendingDelete` restore-callback on the instance, so at most one row confirms at a time — opening a confirm on a second row dismisses the first. Confirm and cancel clicks call `stopPropagation()`, since the row's own click handler opens the session and closes the modal; a leaked click there would be the worst possible misfire. `renderSessionList()` clears the pending callback so a filter change or post-delete re-render can't restore actions onto a detached element, and `onClose()` tears down both the callback and the key listener.
- `src/i18n/en.ts` — two new keys with `context` per the i18n rule (`agent.sessionList.deleteConfirmAction`, `agent.sessionList.deleteCancel`). `agent.sessionList.deleteConfirm` is reused as the prompt; its `context` was updated since it no longer describes a browser dialog. Per-language files regenerate via the `Update UI translations` workflow — no manual translation here, and `t()` falls back to English for the two new keys until it runs.
- `styles.css` — confirm-state styles from the `--gs-*` tokens only. The Delete button reuses the `--gs-error` / `--gs-on-error` pairing the scheduler modal's confirm-delete button uses (`--gs-on-error` landed in #1272, which has since merged, so no competing literal was needed). While confirming, the action area is allowed to shrink and the prompt ellipsizes, with `flex-shrink: 0` on both buttons so the decision controls are never squeezed by a long session title.
- `test/ui/agent-view/session-list-confirm.test.ts` (new) — 9 tests. This modal was on the untested list in #1262, so this is its first coverage.

**Escape handling:** Escape cancels a pending confirm instead of closing the modal, via a capture-phase `keydown` listener on `modalEl` — capture there runs before Obsidian's document-level close handler, so stopping propagation keeps the modal open. Scoping it to `modalEl` rather than `document` avoids swallowing Escape for any other surface. The listener only sees the event when focus is inside the modal, which is the case here because reaching the confirming state requires clicking the row's delete button. If focus were somehow outside, Escape closes the modal — which deletes nothing, so the degradation is safe.

**No docs update.** `docs/` and `README.md` describe session deletion as a user action, not the shape of its confirmation dialog — checked: the only nearby matches are `docs/reference/settings.md:523` (tool-permission confirmations, unrelated) and `README.md:305` (deleting session files manually). Nothing documents the browser dialog, so there is nothing to rewrite.

## Screenshots / Screencast

Not included — no Obsidian runtime in the build sandbox to capture one. See the verification note below; the visual pass is the one thing this PR asks a human to do.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — #1275, plan approved on the issue
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`) — full pre-flight green locally: `format-check`, `lint` (0 errors), `build`, `typecheck:test`, `test` (169 files / 3665 tests), plus `lint:cycles` (no cycles)
- [ ] I have tested this change on Desktop — **not done.** No Obsidian runtime in the build sandbox, so the two-click flow, the Escape path, and the light/dark appearance were **not** exercised end-to-end. Behavioral verification not run in sandbox — needs a manual check. Covered by unit tests at the DOM level, which is not the same thing.
- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards) — **not verified at runtime**, same reason. No platform guard is needed: the change removes a host-chrome dialog and replaces it with plugin-rendered DOM and Obsidian APIs only, which is strictly better on mobile than the `confirm()` it replaces. Worth a tap-target glance on a phone.
- [x] Documentation has been updated (if applicable) — N/A, see "No docs update" above

### AI-Generated Code

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4XsYeBhkuVqWLA18UVDM1)_

## Summary by CodeRabbit

* **New Features**
  * Added inline delete confirmation controls to session list rows, with **Delete** and **Cancel** actions.
  * Escape now cancels an active row confirmation without closing the session list modal.
* **Bug Fixes**
  * Prevented accidental deletion by replacing blocking confirmation dialogs with an inline prompt.
  * Cancel now restores the row’s original actions and avoids cross-row impact.
* **Tests**
  * Added UI coverage for inline confirmation, cancellation, multi-row behavior, and keyboard handling.

### DIFF
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -2744,7 +2744,16 @@ export const en = {
 	},
 	'agent.sessionList.deleteConfirm': {
 		message: 'Delete session "{title}"?',
-		context: 'Browser confirmation dialog before deleting a session. {title} is the session title.',
+		context:
+			'Inline confirmation prompt shown in the session row before deleting a session. {title} is the session title.',
+	},
+	'agent.sessionList.deleteConfirmAction': {
+		message: 'Delete',
+		context: 'Label on the button that confirms deleting a session, next to the inline confirmation prompt.',
+	},
+	'agent.sessionList.deleteCancel': {
+		message: 'Cancel',
+		context: 'Label on the button that dismisses the inline session-delete confirmation without deleting.',
 	},
 	'agent.sessionList.deleted': {
 		message: 'Session "{title}" deleted',
--- a/src/ui/agent-view/session-list-modal.ts
+++ b/src/ui/agent-view/session-list-modal.ts
@@ -22,6 +22,14 @@ export class SessionListModal extends Modal {
 	private projectMap: Map<string, string> = new Map();
 	/** Current filter selection: 'all', 'none', or a project file path. */
 	private selectedFilter: string = FILTER_ALL;
+	/**
+	 * Restores the action buttons of the row currently showing its delete
+	 * confirmation, or null when no row is confirming. Doubles as the "is a
+	 * confirm pending?" flag — at most one row confirms at a time.
+	 */
+	private pendingDelete: (() => void) | null = null;
+	/** Capture-phase Escape handler; cancels a pending confirm before the modal closes. */
+	private escapeHandler: ((evt: KeyboardEvent) => void) | null = null;
 
 	constructor(
 		app: App,
@@ -40,6 +48,17 @@ export class SessionListModal extends Modal {
 		contentEl.empty();
 		contentEl.addClass('gemini-session-modal');
 		this.modalEl.addClass('mod-gemini-session-modal');
+
+		// Escape cancels a pending row confirmation rather than closing the whole
+		// modal. Capture phase on modalEl runs before Obsidian's document-level
+		// close handler, so stopping propagation there keeps the modal open.
+		this.escapeHandler = (evt: KeyboardEvent) => {
+			if (evt.key !== 'Escape' || !this.pendingDelete) return;
+			evt.preventDefault();
+			evt.stopPropagation();
+			this.dismissDeleteConfirm();
+		};
+		this.modalEl.addEventListener('keydown', this.escapeHandler, true);
 
 		// Title
 		contentEl.createEl('h2', { text: t('agent.sessionList.title') });
@@ -172,6 +191,9 @@ export class SessionListModal extends Modal {
 	}
 
 	private renderSessionList(container: HTMLElement) {
+		// Any row mid-confirmation is about to be replaced; drop the stale restore
+		// callback so it can't rebuild actions on a detached element.
+		this.pendingDelete = null;
 		const filtered = this.getFilteredSessions();
 
 		if (filtered.length === 0) {
@@ -222,30 +244,7 @@ export class SessionListModal extends Modal {
 
 			// Actions
 			const actionsDiv = sessionItem.createDiv({ cls: 'gemini-session-actions' });
-
-			// Open button
-			const openBtn = actionsDiv.createEl('button', {
-				cls: 'gemini-session-action-btn',
-				title: t('agent.sessionList.openTooltip'),
-			});
-			setIcon(openBtn, 'arrow-right');
-
-			// Delete button
-			if (this.callbacks.onDelete) {
-				const deleteBtn = actionsDiv.createEl('button', {
-					cls: 'gemini-session-action-btn delete',
-					title: t('agent.sessionList.deleteTooltip'),
-				});
-				setIcon(deleteBtn, 'trash-2');
-
-				deleteBtn.addEventListener('click', (e) => {
-					e.stopPropagation();
-					// eslint-disable-next-line no-alert -- TODO: replace with Obsidian confirmation Modal
-					if (confirm(t('agent.sessionList.deleteConfirm', { title: session.title }))) {
-						void this.deleteSession(session);
-					}
-				});
-			}
+			this.renderSessionActions(actionsDiv, session);
 
 			// Click handler for the entire item
 			sessionItem.addEventListener('click', () => {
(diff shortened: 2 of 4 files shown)


allenhutchison:
Addressed the one actionable finding (`styles.css` — cancel button using raw `--interactive-*`) in 6e2f5a2, though not by the swap that was suggested. The underlying complaint was right; the prescribed fix would have been wrong.

**Why not the literal swap.** There is no `--gs-*` token that aliases `--interactive-normal` / `--interactive-hover`. The token layer's surface tokens are `--gs-surface` (`--background-primary`), `--gs-surface-alt` (`--background-secondary`) and `--gs-hover` (`--background-modifier-hover`) — all different Obsidian variables. Substituting one would be a **value-changing** edit dressed as a token migration, and `docs/contributing/design-system.md` is explicit on both counts: raw vars are the documented holdout for "other Obsidian variables with no `--gs-` equivalent" (alongside the font-size scale and `--size-2-*`), and a value-preserving change must be verified by computed parity **in a live vault** — which this sandbox has no Obsidian runtime for. So the swap would have traded a sanctioned holdout for an unverifiable appearance change.

**What the code does instead.** `ManagementModalBase.confirmDelete()` (`src/ui/components/management-modal-base.ts:163`) already answers this for the same control: its cancel button is a bare `<button>` with **no class at all**, and only the destructive button is styled. So the cancel button's three colour declarations are simply gone — it inherits Obsidian's native button chrome, and sizing still comes from the shared rule it shares with Delete. That removes the raw variables rather than relocating them, adds no token, changes no token semantics, and leaves Delete as the only button in the block reaching for colour (`--gs-error` / `--gs-on-error`). The block is now `--gs-*` throughout apart from `--font-ui-smaller`, which is the documented font-scale holdout. It's also strictly more theme-adaptive than what I had: the button is whatever the active theme makes a button look like.

The 19 other raw `--interactive-normal` / `--interactive-hover` uses elsewhere in `styles.css` are pre-existing and fall under the same documented holdout, so I've left them alone rather than widening this PR — if you'd like tokens minted for interactive-control surfaces, that's a design-system change worth its own issue and its own live-vault parity check.

Pre-flight re-run green: `format-check`, `lint` (0 errors), `build`, `typecheck:test`, `test` (169 files / 3665 tests). CSS-only change; the 9 new tests assert behaviour, not colour, and still pass.

One note on the ast-grep warning for `document.body.innerHTML = ''` in the new test: that's a constant empty string in `beforeEach` teardown, matching the existing convention in `test/ui/update-notification-modal.test.ts` and others. No untrusted input is involved, so I've left it consistent with its neighbours rather than making this one file an outlier.

**Still unverified, deliberately:** no Obsidian runtime here, so the cancel button's native appearance in light/dark — like the rest of this PR's visual surface — hasn't been seen running. The Desktop/Mobile checklist items stay unticked.

---

allenhutchison:
On the follow-up note asking to verify the inherited Cancel styling in a live vault across Desktop/Mobile and light/dark: agreed, and that check is not something this pipeline can perform — the build sandbox has no Obsidian runtime. It is already recorded as such rather than quietly assumed: the Desktop and Mobile checklist items above are deliberately unticked, and the PR body carries `behavioral verification not run in sandbox`. The same gap covers the whole visual surface of this change, not just the Cancel button.

Worth noting the direction of the risk here, since the guideline quoted is about token migrations. This change removed colour declarations rather than substituting a token, so Cancel now renders with Obsidian's own button chrome — the same styling every unclassed `<button>` in the plugin's modals gets, including the cancel button in `ManagementModalBase.confirmDelete()`. If it were illegible in some theme, every stock button in that theme would be too. That makes it a lower-risk state than the `--interactive-*` literals it replaced, though it is still unverified, which is why the checklist stays honest about it.

CI is green on 6e2f5a2 (test, lint, CodeQL, both Analyze jobs, dead-code report). Nothing further from me on this PR — it's ready for your review, with the one-vault visual pass as the outstanding item.

---

Issue #1252, opened 2026-07-25
Ollama Ignores Vault Context


SilasCodesGames:
I've installed Ollama and I'm running gemma4. 
I've confirmed that Gemini Scribe can connect to Ollama, but every request I send gives un-relevant answers. When I expand the "Reasoning" in the Agent Session chat, I see "The user wants to know what they should do with a Project Alpha project. I need to find information about Project alpha. I'll start by searching..."

And then the agent starts searching for files that don't exist.

I've reproduced this behavior in multiple vaults. This looks like Gemini Scribe agent is falling back to some demo / default vault context.

allenhutchison:
Thanks for the detailed report, @SilasCodesGames. The reasoning trace you quoted — the agent inventing a "Project Alpha" and then searching for files that don't exist — is the useful clue here. That string isn't a hardcoded default anywhere in the plugin (it only appears in the repo's eval fixtures, never in shipped code), so this looks like either (a) the vault context isn't reaching the Ollama request the way it does for Gemini, or (b) the specific model is too small to follow the agent's tool-use instructions and is confabulating. A few specifics would let the next pass tell those apart and plan a fix:

1. **Plugin version** — Settings → Community plugins → Gemini Scribe (or the `version` in the plugin's `manifest.json`).
2. **Exact Ollama model tag** — "gemma4" isn't a standard tag; is it `gemma4:e4b`, `gemma3:4b`, `gemma2`, or something else? (`ollama list` shows the exact name.) Small models often can't drive the agent's tool loop reliably, so the exact tag matters.
3. **Does the same vault + prompt work when you switch the provider to Gemini?** This is the key isolation step — if Gemini answers correctly in the same vault but Ollama doesn't, it points at Ollama-specific context handling; if both misbehave, it points at the vault/context setup.
4. **Debug logs** — enable Debug mode (Settings → Gemini Scribe → Advanced/Debug), reproduce once, and open the developer console (Ctrl/Cmd+Shift+I). Do the `[Gemini Scribe]` logs show your actual note contents / file list being included in the request sent to Ollama, or is that context empty/missing?
5. **What is "Project Alpha"?** Is there any note in your vault that mentions it, or is the model inventing it entirely from nothing?

---

Reply here and the next automation pass will pick it up, or add the Skip label to opt this issue out of automation.

---
_Generated by [Claude Code](https://claude.ai/code)_

allenhutchison:
[...]

Issue #1237, opened 2026-07-22
feat: Support ChatGPT-compatible integration


evilmarty:
The Ollama integration is great for running local modals but it does require the use of Ollama. Alternative runtimes such as LM Studio and oMLX etc support interoperability via a ChatGPT-compatible API, which Ollama also supports. It would be great to add support so a wider range of integrations are available to be used by this plugin.

allenhutchison:
Thanks for the request, @evilmarty. An OpenAI-/ChatGPT-compatible provider is very much in scope conceptually — the API layer already has a provider abstraction (`ModelClientFactory` branches on `settings.provider` to build a `GeminiClient` or `OllamaClient`, each under `src/api/providers/{gemini,ollama}/`), so a third `openai-compatible` provider that talks to LM Studio / MLX / Ollama's own OpenAI-compatible endpoint would slot into that shape.

This isn't blocked on missing detail — it's waiting on a call that's yours to make: how to sequence it against the provider work that's already open. There's an active **Provider abstraction epic (#703)**, and the related provider-expansion issues — **#704** (per-use-case provider selection), **#705** (Ollama embeddings/RAG), **#706** (Ollama image-gen parity) — are all currently **parked**. A standalone OpenAI-compatible provider could either land now against today's abstraction or fold into that epic, and whether to widen the provider surface while the rest of it is on hold is a roadmap decision the pipeline shouldn't make for you.

Want me to **park** this until you've decided how it fits the provider roadmap? Reply "park it" (or add the Parked label) and I'll leave it untouched until you remove the label or add more detail. If you'd rather move it forward now, reply "go ahead" (optionally with the scope you want — e.g. "standalone `openai-compatible` provider, chat + agent only, no RAG/image-gen") and the next pass will draft a full implementation plan.

---

Reply here and the next automation pass will pick it up, or add the Skip label to opt this issue out of automation.

---
_Generated by [Claude Code](https://claude.ai/code)_

evilmarty:
@allenhutchison I've seen the abstraction factory issue and I agree it's worth pursuing but I don't have a plan of attack for that implementation just yet. I believe adding the ChatGPT-compatibile provider would give more perspective on how best to approach that bit of work. If you have no issue, I am keen to add that bit of work and then look at refining it with better abstraction.

allenhutchison:
Thanks, @evilmarty — you offering to implement this changes the picture, and the call is now @allenhutchison's to make.

[...]

PR #1288, opened 2026-07-31
fix(openai): make the GPT-5.6 models usable — unblock 400s, trim the catalog, fix context windows

allenhutchison:
## Summary

The OpenAI provider merged in #1286 could not complete a single request against the GPT-5.6 models on `api.openai.com` — every chat turn failed with a 400. This fixes that, trims the model dropdowns to the models the client can actually drive, and fixes a context-window bug found while chasing the first two.

Follow-up to #1286 / #1237. There is no separate issue for this; it was found by installing the merged build into a real vault and using it.

## Changes

- **`gpt-5.6*` requests no longer 400.** Two stacked causes: the client always sent `temperature`/`top_p`, which these reasoning models reject outright, and `/v1/chat/completions` rejects function tools for them unless `reasoning_effort` is `'none'` — so every agent-mode turn failed. Sampling params are now omitted for these models and tool-carrying requests pin `reasoning_effort: 'none'`. Gated on model id, not base URL, so a proxy serving them behaves the same; all other models are untouched.
- **API errors are now readable.** The console serialized the SDK's nested error body as `"error":"Object","code":"null"`, which hid the server message. This is why the temperature failure was invisible until reproduced by hand. Errors now log as `HTTP 400 [unsupported_value]: Unsupported value: 'temperature'…`.
- **Hosted model list narrowed to three.** `api.openai.com` advertises ~90 ids, most undrivable through Chat Completions; the old regex still let 91 through into every dropdown. Now an allowlist of `gpt-5.6-sol` (chat), `gpt-5.6-terra` (summary), `gpt-5.6-luna` (completions). Custom base URLs are unaffected — a compatible server's catalog is still taken at face value.
- **Context windows corrected to 922,000** for all three. `/v1/models` returns no capability metadata whatsoever, so these are curated by hand, and all three curated values were wrong (1M / 400k / 128k).
- **The context manager now reads a model's own window.** `getInputTokenLimit` consulted per-model data only for Ollama; every other provider got the flat provider default, so `contextWindow` was dead data for OpenAI. This is #1252's bug in the other direction — the 128k floor against a 922k model meant history was compacted at ~14% of real capacity.

## How the 922,000 figure was determined

Not from documentation — `/v1/models` returns only `id`, `created`, and `owned_by`. It was measured against the live API: an over-limit request is rejected before inference and names the ceiling in its error.

```
Input tokens exceed the configured limit of 922000 tokens.
Your messages resulted in 1120008 tokens.
```

Same result for `sol` and `luna`; `terra` accepted 800k, well past its claimed 400k. That procedure is documented in a code comment so the values can be re-measured when the family changes, rather than re-guessed.

## Reviewer notes

- The three commits are independently reviewable, and **the third is not OpenAI-specific** — it changes limit resolution for any provider whose models carry per-model windows. It is safe as written: Gemini's bundled entries declare no `contextWindow`, so they keep the 1M provider default, and Ollama's more authoritative runtime-allocation path is untouched. Happy to split it into its own PR if you would rather land it separately.
- The corrected window values in commit 2 are inert until commit 3 — worth knowing if you review them in isolation.
- `reasoning_effort: 'none'` means agent-mode turns on GPT-5.6 run without reasoning output. Tool calling with reasoning would require the Responses API, which this client deliberately does not target.

## Screenshots / Screencast

No UI changes. The one user-visible difference is the token budget readout, verified in a live vault: the same context previously reported `~10,657 / 128,000 (8.3%)` and now correctly reports `~10,657 / 922,000 (1.2%)`.

### Required

- [ ] This PR is linked to an approved issue where the approach was discussed with a maintainer — **no prior issue**; this is a defect found while testing the merged #1286 in a real vault
- [x] All CI checks pass (`npm test` 3789 passing, `npm run build`, `npm run format-check`, `npm run typecheck:test`, `npm run lint`, `npm run lint:cycles`)
- [x] I have tested this change on Desktop — installed into a live vault; agent turn with tool calling verified end to end against the real API
- [x] I have verified this change does not break Mobile — provider/plumbing logic only; no platform APIs or Node built-ins introduced
- [x] Documentation has been updated — `docs/guide/openai-setup.md`

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Added support for validated GPT-5.6 models with curated context limits and metadata.
  * Improved context-limit detection using discovered model information.
  * Compatible OpenAI endpoints can use their available model catalogs with fallback metadata.
* **Bug Fixes**
  * GPT-5.6 requests now handle sampling and tool-calling constraints correctly.
  * API and streaming errors provide clearer status, code, and message details.
* **Documentation**
  * Updated OpenAI setup guidance with model filtering, endpoint behavior, and GPT-5.6 limitations.

### DIFF
--- a/docs/guide/openai-setup.md
+++ b/docs/guide/openai-setup.md
@@ -26,12 +26,20 @@ Example with [LM Studio](https://lmstudio.ai/):
 
 The same pattern works for MLX-served endpoints, Ollama's OpenAI-compatible endpoint (`http://localhost:11434/v1`), or any other server that implements the `/v1/chat/completions` and `/v1/models` endpoints.
 
-Requests to a non-`api.openai.com` base URL are not filtered against OpenAI's own non-chat model ids (embeddings, TTS, etc.) — a compatible server's `/models` catalog is taken at face value.
+A compatible server's `/models` catalog is taken at face value — the supported-model allowlist described below applies only to `api.openai.com`.
 
 ## Models
 
-- **`api.openai.com`** — the model list is enriched with curated metadata (context window, vision support) for current OpenAI models. `gpt-5.6` (chat), `gpt-5.6-terra` (summary), and `gpt-5.6-luna` (completions) are the defaults; older `gpt-5.x` and `gpt-4.x` models remain selectable.
+- **`api.openai.com`** — the endpoint advertises around ninety model ids, most of which this integration can't drive usefully (Responses-API-only, audio, embeddings, older families). The dropdowns are therefore limited to the three GPT-5.6 models that have been validated against the plugin: **`gpt-5.6-sol`** (chat default), **`gpt-5.6-terra`** (summary default), and **`gpt-5.6-luna`** (completions default). All three accept **922,000 input tokens**, which is what the token counter above the message box reflects.
+- **Where those numbers come from** — OpenAI's `/v1/models` returns no capability metadata (just an id and a creation date), so context windows are curated in the plugin rather than discovered. The 922,000 figure was measured directly against the live API, not taken from documentation.
 - **Compatible servers** — an unrecognized model id gets a conservative default: 128k context window, no vision, tool calling assumed. This covers most locally-served models correctly for tool calling, but vision won't be enabled automatically — there's no metadata to detect it from.
+
+### Reasoning and tool calling on GPT-5.6
+
+The GPT-5.6 models are reasoning models, which constrains what Chat Completions accepts:
+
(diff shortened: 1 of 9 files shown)


allenhutchison:
Addressed the `describeError` nitpick in 0f35f0f: the formatted string and the original error are now both passed to `logger.error`, so structured API bodies stay readable while ordinary errors keep their stack trace. Added a regression test asserting the raw error is forwarded.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1434, opened 2026-08-31
fix(attachments): hold the 20 MB budget against the rasterized PNG everywhere

### LINKED ISSUE #1430: invariant drift: the post-rasterize attachment budget check landed at one of three SVG call sites
## What

The `GEMINI_INLINE_DATA_LIMIT` (20 MB) budget is supposed to bound what actually gets sent to the model. For SVG that is a subtle rule, because the plugin never sends the SVG — it rasterizes to PNG first, and a decoded bitmap can be **orders of magnitude larger than its source**. A 40 KB SVG that declares `width="20000" height="20000"` rasterizes to a several-hundred-megabyte PNG.

The #1412 review caught exactly this and added a post-rasterize re-check in `attach-vault-file.ts`:

```ts
// The rasterized PNG can be larger than the source SVG (it's a decoded
// bitmap), so the budget — checked above against the small source file —
// has to hold for the converted payload too (#1412 review).
const convertedBytes = base64DecodedBytes(base64);
if (alreadyUsedBytes + convertedBytes > GEMINI_INLINE_DATA_LIMIT) {
    return { kind: 'too-large' };
}
```

That fix landed in the shared leaf only. There are **three** places that call `rasterizeSvg` and then hand the PNG onward, and the other two still check the budget against the *source* size and never look at the converted payload. So the policy is correct in one third of the code that implements it, and the two remaining copies are wrong in the way the review already identified.

This is drift in the shape this repo keeps producing: a review fixes the call site in front of it, the siblings implementing the same policy aren't grepped for, and the fix doesn't propagate. It is not visible to `jscpd` — the three sites are not textual clones (one is a tool result, one is a UI counter loop, one is the leaf), so #1264's duplicate-block detector would not catch it either.

Routed as an issue rather than a PR because it spans three modules, changes user-visible behavior (files that attach today would start being rejected), and the right shape for the shared helper is a design call — see **Proposed unit of work**.

## Evidence

**1. `src/tools/vault/read-file-tool.ts:124–145` — the agent's autonomous `read_file` on an SVG**

```ts
const buffer = await plugin.app.vault.readBinary(file);
if (buffer.byteLength > GEMINI_INLINE_DATA_LIMIT) {   // ← source size
    return { success: false, error: `File too large for inline processing (max 20 MB): ${file.name}` };
}
try {
    const base64 = await rasterizeSvg(buffer, file.extension.toLowerCase() === 'svgz');
    return {
        success: true,
        data: { path: file.path, type: 'binary_file', mimeType: 'image/png', size: buffer.byteLength },
        inlineData: [{ base64, mimeType: 'image/png' }],   // ← no re-check
    };
```

No post-rasterize check at all. The oversized PNG goes straight into `inlineData` and is injected into conversation history by the tool pipeline. Secondary defect in the same block: `data.size` reports `buffer.byteLength`, the *source* SVG size, while `inlineData` carries the much larger PNG — so the size the model is told about is not the size of the payload it received.

**2. `src/ui/agent-view/agent-view-ui.ts:873–890` + `:949–961` — the external-drop / clipboard-paste path**

`attachExternalSvgFile` rasterizes and calls `callbacks.addAttachment(attachment)` with no size check of any kind. Its caller `processAttachmentFiles` gates on the source `File.size`:

```ts
} else if (this.isSvgFile(file)) {
    if (cumulativeSize + file.size > GEMINI_INLINE_DATA_LIMIT) {   // ← source size
        new Notice(t('agent.attachments.sizeLimitReached'));
        break;
    }
    notifyFirstImage();
    if (await this.attachExternalSvgFile(file, callbacks)) {
        cumulativeSize += file.size;   // ← source size again
```

Two distinct problems here. The gate uses the source size, same as above. But `cumulativeSize += file.size` also **understates the running total**, and that error compounds: drop five SVGs and the counter believes it has consumed the sum of five small XML files while the attachment list holds five large bitmaps. Every subsequent attachment — image, SVG, or otherwise — is then measured against a budget that has already been silently overspent. Note the seed is correct (`getCurrentAttachmentSize` → `estimateAttachmentBytes`, which decodes base64); only the in-loop increment is wrong, so the understatement is repaired on the next fresh call and reappears within each one.

**3. `src/ui/agent-view/attach-vault-file.ts:72–90` — the vault-drop / @-mention path (correct)**

The shared leaf from #1363/#1412. Checks the source size for the read budget, then re-checks `base64DecodedBytes(base64)` after rasterizing and returns `{ kind: 'too-large' }` when the converted payload blows the budget. Returns `bytes` as the *converted* count so callers accumulate the real payload size. This is the reference implementation.

Consequence when the guard is missed: an over-budget `inlineData` part reaches the request builder, and the failure surfaces as an opaque API-side rejection (or a very slow request) rather than the local "too large" notice the user gets on the vault path today. It also fires on ordinary content — a legitimately detailed vector diagram at a large canvas size, not just a crafted file.

## Proposed unit of work

- Decide where the shared "rasterize, then hold the budget against the converted payload" step lives. Two reasonable shapes: (a) fold the re-check into `rasterizeSvg` itself by giving it a byte budget parameter and letting it throw/return a too-large result — every caller then gets it for free; (b) extract a small `rasterizeSvgWithinBudget(buffer, isSvgz, alreadyUsedBytes)` leaf next to `svg-rasterizer.ts` and route all three sites through it. (a) is fewer call-site edits; (b) keeps `rasterizeSvg` a pure converter. Pick one before touching the call sites.
- Route `read-file-tool.ts`'s SVG branch through it, and return an error result rather than oversized `inlineData` when the converted PNG exceeds the limit. Wording should match the existing `File too large for inline processing (max 20 MB): ${file.name}` string, which is model-facing and stays English.
- Fix `read-file-tool.ts`'s `data.size` to report the size of the payload actually returned (the decoded PNG), not the source SVG. Do this in the same change — it is the same mistake, and a model reasoning about the result reads that field.
- Route `attachExternalSvgFile` through it, and change `processAttachmentFiles` to add the *converted* byte count to `cumulativeSize` rather than `file.size`. Have `attachExternalSvgFile` return the byte count (or a discriminated result mirroring `VaultAttachmentResult`) instead of a bare boolean, so the caller has the real number to add.
- While in `processAttachmentFiles`: the non-SVG image branch adds `file.size` too. That one is close enough (base64 of the raw bytes decodes back to `file.size`) that it is defensible, but if the SVG branch starts returning real byte counts, consider making both branches use the same source of truth so the next reader doesn't have to re-derive why one is exact and one is an estimate.
- Cover with tests: `test/tools/vault/` for the tool path (a stub `rasterizeSvg` returning an oversized base64 → assert an error result, not `inlineData`), and `test/ui/agent-view-ui.test.ts` for the cumulative counter (drop two SVGs whose rasterized output exceeds the budget → assert the second is rejected). `test/utils/svg-rasterizer.test.ts` already exists for the converter itself.

## Out of scope

- The duplication of the read → classify → base64 → webm-sniff pipeline across these same three modules. That is real, and `read-file-tool.ts`'s `GEMINI_BINARY` branch is a fourth partial copy, but unifying the whole pipeline is a larger refactor with a different rationale — this issue is only about the budget policy being wrong in two places. File that separately if it is worth doing.
- Changing `GEMINI_INLINE_DATA_LIMIT` itself, or introducing a separate (lower) cap for rasterized output. Arguably a rasterized PNG deserves a tighter bound than a natively-supported 20 MB video, but that is a product decision, not this fix.
- Bounding the rasterizer's output dimensions (clamping a 20000×20000 canvas before it is allocated) rather than measuring after the fact. That is the more robust guard — the current fix still allocates the bitmap before rejecting it — but it is a change to `svg-rasterizer.ts`'s contract and worth its own discussion.
- The `read-file-tool.ts` `GEMINI_BINARY` branch's budget check, which is correct as written (base64 of raw bytes, no conversion, so source size is the payload size).

---

allenhutchison:
## Summary

Fixes #1430

The 20 MB inline-data budget for SVG must hold against the **rasterized PNG**, not the source file — a decoded bitmap can dwarf a small SVG. The #1412 review added that post-rasterize check in one of three call sites (`attach-vault-file.ts`); the other two — `read-file-tool.ts` (the agent's autonomous `read_file`) and the external-drop/paste path in `agent-view-ui.ts` — still gated on the *source* size (or nothing) and sent over-budget PNGs onward, surfacing as opaque API rejections or very slow requests. The drop path also understated the cumulative attachment counter (`cumulativeSize += file.size` while the shelf held large bitmaps), compounding across multiple SVGs so every later attachment was measured against an already-overspent budget.

**Design (the issue's fork, option a):** the budget check now lives inside `rasterizeSvg` itself — a new optional `budgetBytes` parameter throws a typed `SvgTooLargeError` when the decoded PNG exceeds it. Every current and future caller gets the guard for free; the alternative (a separate `rasterizeSvgWithinBudget` leaf) would keep `rasterizeSvg` callable bare and re-open the exact drift surface this issue exists to close.

## Changes

- `src/utils/svg-rasterizer.ts` — `rasterizeSvg(buffer, isSvgz, budgetBytes?)` + exported `SvgTooLargeError` (carries the decoded byte count). Behavior is unchanged when `budgetBytes` is omitted.
- `src/utils/file-classification.ts` — `base64DecodedBytes` moved here from `inline-attachment.ts` (this leaf already owns `GEMINI_INLINE_DATA_LIMIT` and `arrayBufferToBase64`, so the rasterizer can use the formula without importing a UI module); `inline-attachment.ts` re-exports it so every existing import path keeps working.
- `src/tools/vault/read-file-tool.ts` — SVG branch passes the full limit as budget; `SvgTooLargeError` maps to the existing model-facing `File too large for inline processing (max 20 MB)` error string; `data.size` now reports the decoded PNG size (previously it reported the source SVG size while `inlineData` carried the much larger PNG — the model was misinformed about the payload it received).
- `src/ui/agent-view/agent-view-ui.ts` — `attachExternalSvgFile(file, callbacks, budgetBytes)` returns the decoded byte count on success, `'too-large'`, or `null` (raster failure); `processAttachmentFiles` passes the remaining budget and adds **real payload bytes** to the running total in both the SVG *and* plain-image branches (the issue's "consider" item — one formula governs the counter, so the next reader doesn't re-derive why one branch was exact and the other an estimate).
- `src/ui/agent-view/attach-vault-file.ts` — the shared pipeline passes `GEMINI_INLINE_DATA_LIMIT − alreadyUsedBytes` into `rasterizeSvg` and maps `SvgTooLargeError` to its existing `'too-large'` result; the inline post-check collapses away. `VaultAttachmentResult` is unchanged for callers.
- `docs/guide/agent-mode.md` — one sentence in the SVG Handling section documenting that the budget applies to the converted PNG and that over-budget SVGs are rejected with the size-limit notice.
- Tests: `svg-rasterizer.test.ts` (budget throw / boundary-inclusive fit / unbounded when omitted), `read-file-tool.test.ts` (oversized rasterization → error result not `inlineData`; `data.size` = decoded; raster failure path), `agent-view-ui.test.ts` (delegate receives the remaining budget; running total holds converted bytes so a second 9 MB SVG after a 12 MB one is rejected; `'too-large'` breaks with the size notice), `attach-vault-file.test.ts` (typed error maps to `'too-large'`; budget passed through to the converter).

## User-visible impact

SVGs whose rasterized output would exceed 20 MB are now rejected locally with the size-limit notice on **every** path (previously: silently over-budget on `read_file` and external drop/paste). Attachments at legitimate sizes are unaffected.

## Screenshots / Screencast

N/A — behavioral guard; the change surfaces as an existing localized notice, already covered by the docs update.

### Required

### AI-Generated Code

Notes on the unticked items:

- **Desktop testing**: not performed — the budget math and every conversion path are exercised by the test suite (3923 tests green); the rasterizer itself runs on the same `<canvas>` code as before, only with a post-conversion size comparison added. A manual desktop pass would only re-confirm the same notice text already asserted in tests.
- **Mobile**: no platform-specific API touched; the rasterizer's canvas path is unchanged. The zlib `no-nodejs-modules` lint warning in `test/utils/svg-rasterizer.test.ts` is pre-existing (test-only, gated off in the `test/**` ESLint override).
- **Documentation**: `docs/guide/agent-mode.md` SVG Handling section updated — the budget-now-applies-to-the-converted-PNG rule is user-visible behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * SVG attachments now enforce the 20 MB limit against decoded rasterized PNG size.
  * Oversized SVG conversions display the standard file-size limit notice.
  * Inline attachment tracking accurately accounts for converted SVG output and cumulative limits.

* **Documentation**
  * Clarified that the 20 MB limit applies to rasterized PNG output rather than the original SVG file.

### DIFF
--- a/docs/guide/agent-mode.md
+++ b/docs/guide/agent-mode.md
@@ -107,7 +107,7 @@ When you drop a file, the plugin classifies it based on its extension:
 
 **SVG Handling:**
 
-Gemini's API can't process `image/svg+xml` directly, so SVG (and gzip-compressed `.svgz`) files are **rasterized to PNG on your device** before being sent — whether you drag, paste, `@`-mention, or have the agent read them with the Read File tool. Rasterization renders the SVG onto a white background (so transparent artwork like handwritten ink strokes stays legible for OCR) and caps the longest edge at 2048px to keep the payload within the inline-data limit. If an SVG can't be rendered (malformed markup or unresolvable external references), it's skipped with the same "unsupported file type" notice rather than sending unusable data.
+Gemini's API can't process `image/svg+xml` directly, so SVG (and gzip-compressed `.svgz`) files are **rasterized to PNG on your device** before being sent — whether you drag, paste, `@`-mention, or have the agent read them with the Read File tool. Rasterization renders the SVG onto a white background (so transparent artwork like handwritten ink strokes stays legible for OCR) and caps the longest edge at 2048px to keep the payload within the inline-data limit. The rasterized PNG is a decoded bitmap, so it can be far larger than the source SVG; the 20 MB inline-data budget applies to the converted PNG, and an SVG whose rasterized output would exceed it is rejected with the same size-limit notice as any other oversized file. If an SVG can't be rendered (malformed markup or unresolvable external references), it's skipped with the same "unsupported file type" notice rather than sending unusable data.
 
 **How It Works:**
 
--- a/src/tools/vault/read-file-tool.ts
+++ b/src/tools/vault/read-file-tool.ts
@@ -9,9 +9,10 @@ import {
 	FileCategory,
 	GEMINI_INLINE_DATA_LIMIT,
 	arrayBufferToBase64,
+	base64DecodedBytes,
 	detectWebmMimeType,
 } from '../../utils/file-classification';
-import { rasterizeSvg } from '../../utils/svg-rasterizer';
+import { rasterizeSvg, SvgTooLargeError } from '../../utils/svg-rasterizer';
 import { resolvePathToFileOrFolder, toFileEntry } from './utils';
 
 /**
@@ -130,13 +131,26 @@ export class ReadFileTool implements Tool {
 					return { success: false, error: `File too large for inline processing (max 20 MB): ${file.name}` };
 				}
 				try {
-					const base64 = await rasterizeSvg(buffer, file.extension.toLowerCase() === 'svgz');
+					// The rasterized PNG is a decoded bitmap that can dwarf the source
+					// SVG, so the budget holds for the converted payload too (#1430).
+					const base64 = await rasterizeSvg(buffer, file.extension.toLowerCase() === 'svgz', GEMINI_INLINE_DATA_LIMIT);
 					return {
 						success: true,
-						data: { path: file.path, type: 'binary_file', mimeType: 'image/png', size: buffer.byteLength },
+						data: {
+							path: file.path,
+							type: 'binary_file',
+							mimeType: 'image/png',
+							size: base64DecodedBytes(base64),
+						},
 						inlineData: [{ base64, mimeType: 'image/png' }],
 					};
 				} catch (rasterErr) {
+					if (rasterErr instanceof SvgTooLargeError) {
+						return {
+							success: false,
+							error: `File too large for inline processing (max 20 MB): ${file.name}`,
+						};
+					}
 					return {
 						success: false,
 						error: `Failed to rasterize SVG for viewing: ${file.name} (${getRawErrorMessageOr(rasterErr, 'Unknown error')})`,
--- a/src/tools/vault/read-file-tool.ts
+++ b/src/tools/vault/read-file-tool.ts
@@ -127,12 +127,10 @@ export class ReadFileTool implements Tool {
 				// to PNG so the agent can actually view/OCR it. On failure, return an error
 				// string rather than sending anything unusable to the API.
 				const buffer = await plugin.app.vault.readBinary(file);
-				if (buffer.byteLength > GEMINI_INLINE_DATA_LIMIT) {
-					return { success: false, error: `File too large for inline processing (max 20 MB): ${file.name}` };
-				}
 				try {
-					// The rasterized PNG is a decoded bitmap that can dwarf the source
-					// SVG, so the budget holds for the converted payload too (#1430).
+					// No source-byte gate: the payload actually sent is the rasterized
+					// PNG, which can be far smaller than a bulky source SVG. The budget
+					// holds for the converted payload, enforced inside rasterizeSvg (#1430).
 					const base64 = await rasterizeSvg(buffer, file.extension.toLowerCase() === 'svgz', GEMINI_INLINE_DATA_LIMIT);
 					return {
 						success: true,
--- a/src/ui/agent-view/agent-view-ui.ts
+++ b/src/ui/agent-view/agent-view-ui.ts
@@ -7,6 +7,7 @@ import { renameSessionHistoryFile } from '../../agent/session-rename';
 import { collectFilesFromFolder } from '../../utils/folder-walk';
 import {
 	InlineAttachment,
+	base64DecodedBytes,
 	estimateAttachmentBytes,
 	fileToBase64,
 	generateAttachmentId,
@@ -15,7 +16,7 @@ import {
 } from './inline-attachment';
 import { attachVaultBinaryFile } from './attach-vault-file';
 import { classifyFile, FileCategory, GEMINI_INLINE_DATA_LIMIT } from '../../utils/file-classification';
-import { rasterizeSvg } from '../../utils/svg-rasterizer';
+import { rasterizeSvg, SvgTooLargeError } from '../../utils/svg-rasterizer';
 import { t } from '../../i18n';
 
 /**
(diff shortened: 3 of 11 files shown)


allenhutchison:
Round 2 — outside-diff finding (attach-vault-file.ts:64, source-size gate) addressed in 507b58b.

**Verified valid against current code.** Both `attachVaultBinaryFile` and the `read_file` SVG branch rejected on the raw source buffer before `rasterizeSvg` could measure the PNG actually sent — a >20 MB source SVG that rasterizes small (2048px cap) was refused on source size alone.

Fix as suggested: classification now precedes the gate. Regular binaries keep the source-byte check (base64 of raw bytes decodes back to source size, so source size is the payload size); SVGs skip it and rely solely on the converted-payload budget inside `rasterizeSvg`, which `#1430` added for the inverse case (small source, huge PNG). Regression tests in both suites pin the new behavior: a 21 MB source SVG rasterizing to 1 decoded byte is accepted, with the full remaining budget passed as the converter ceiling.

Full suite 3925/3925, build, typecheck, lint, cycles, knip all green.

PR #1439, opened 2026-09-01
refactor(agent): derive tool-sort priority from tool.classification

### LINKED ISSUE #1424: weak abstraction: TOOL_PRIORITY re-declares by hand the classification every Tool already carries
## What

The reads-before-writes invariant is enforced by `sortToolCallsByPriority` in `src/agent/agent-loop-helpers.ts`, which sorts a batch using `TOOL_PRIORITY` — a hand-maintained, string-keyed map of all 23 tool names to a number, grouped into four bands whose comments are literally `READS (1–19)`, `EXTERNAL (20–29)`, `WRITES (30–39)`, `DESTRUCTIVE (40+)`.

Those four bands are the four values of `ToolClassification` (`src/types/tool-policy.ts`), and `Tool.classification` is a **required** member of the tool interface (`src/tools/types.ts:110`) that every one of the 23 tools already declares. So the map is a second, parallel copy of a fact the tool objects already state — and unlike the declaration, the map is keyed by string, so nothing compile-checks it.

The map is currently in full agreement with the declarations (I checked all 23), so this is latent drift rather than a live bug. But the failure mode is specific and lands on a load-bearing invariant: a tool added without a `TOOL_PRIORITY` entry silently falls to `UNKNOWN_TOOL_PRIORITY = 29`, the **bottom of the EXTERNAL band** — which sorts it *before* `write_file` (30) and `delete_file` (41). For a new READ or EXTERNAL tool that is correct; for a new WRITE or DESTRUCTIVE tool it is exactly the race the sort exists to prevent. The source comment acknowledges the gap rather than closing it:

> Unknown tools fall to the END of the EXTERNAL band (29) — safer than after deletes, since most unknown tools added later will be reads or writes, and if it really is destructive the explicit entry should be added.

"the explicit entry should be added" is the same uncompile-checked convention that #1315 and #1308 were filed for.

## Evidence

- `src/agent/agent-loop-helpers.ts:38–66` — `TOOL_PRIORITY`, 23 hand-written entries, band comments restating `ToolClassification`
- `src/agent/agent-loop-helpers.ts:74` — `UNKNOWN_TOOL_PRIORITY = 29`, the silent fallback for an unlisted tool
- `src/agent/agent-loop-helpers.ts:80–86` — `sortToolCallsByPriority<T extends { name: string }>()`, which only ever sees the call's name
- `src/tools/types.ts:110` — `classification: ToolClassification` required on every tool
- `src/types/tool-policy.ts:26–35` — `READ` / `WRITE` / `DESTRUCTIVE` / `EXTERNAL`, the same four bands
- `src/tools/tool-registry.ts:82` — the registry already resolves `tool.classification` by name for the permission system, so the name → classification lookup exists
- Sole caller: `src/agent/agent-loop.ts:347`

## Proposed unit of work

- Derive the band from the tool's declared `classification` instead of from the name. The blocker is that `agent-loop-helpers.ts` is deliberately pure and UI-agnostic — it takes `{ name: string }[]` and has no registry handle — so the resolution has to be injected, not imported.
- Add a small `classificationToPriority(c: ToolClassification): number` mapping in a leaf module (the four band bases), and change `sortToolCallsByPriority` to take a resolver: `sortToolCallsByPriority(calls, resolve: (name: string) => ToolClassification | undefined)`. `AgentLoop` already holds the tool registry, so it can pass `(n) => registry.getTool(n)?.classification` at `agent-loop.ts:347`.
- Keep an explicit fallback for a name the registry can't resolve, and make it louder than today: `logger.warn` on the unknown name so a missing registration surfaces instead of silently sorting at 29.
- Decide what to do with intra-band ordering. Today the map imposes a total order within a band (`read_file` before `list_files`, etc.); a classification-derived sort makes the whole band equal and relies on the sort's stability to keep the model's emitted order. That is a real behavior change worth naming explicitly in the PR — if any intra-band order is load-bearing, keep a small tiebreak map for just those tools rather than all 23.
- Extend `test/agent/agent-loop-helpers.test.ts`: a DESTRUCTIVE tool that is *not* in any hardcoded list still sorts after every write, which is the case today's map cannot express.

## Out of scope

Not changing `ToolClassification`'s values or the permission system that consumes them. Not touching `ToolCategory` (`src/types/agent.ts`), which is a separate axis. Not revisiting the loop-detection or turn-budget logic that shares this file.

---

allenhutchison:
## Summary

Fixes #1424

`sortToolCallsByPriority` enforced the reads-before-writes invariant via `TOOL_PRIORITY` — a hand-maintained, string-keyed map of all 23 tool names whose four band comments (`READS 1–19`, `EXTERNAL 20–29`, `WRITES 30–39`, `DESTRUCTIVE 40+`) restated the four `ToolClassification` values every tool already declares. Nothing compile-checked the map: a tool added without an entry silently fell to `UNKNOWN_TOOL_PRIORITY = 29` — the bottom of the EXTERNAL band — sorting a new WRITE or DESTRUCTIVE tool *before* `write_file` (30) and `delete_file` (41), exactly the race the sort exists to prevent. The source comment acknowledged the gap ("the explicit entry should be added"); this makes the gap unexpressable instead.

## Changes

- `src/agent/agent-loop-helpers.ts` — `TOOL_PRIORITY` (23 entries) and `UNKNOWN_TOOL_PRIORITY` deleted. New exported `classificationToPriority(c)` maps the four bands (READ=10, EXTERNAL=20, WRITE=30, DESTRUCTIVE=40); `sortToolCallsByPriority(calls, resolve)` takes an injected `name → ToolClassification | undefined` resolver, keeping the module pure and registry-free (the issue's named constraint). Unresolvable names keep the old conservative fallback — end of the EXTERNAL band, after all known reads, before any known write.
- `src/agent/agent-loop.ts` — the sole caller passes `plugin.toolRegistry.getTool(name)?.classification` and `logger.warn`s any name the registry can't resolve, so a missing registration surfaces instead of silently mis-sorting. This is the "louder than today" fallback the issue asked for.
- Tests:
  - `test/agent/agent-loop-helpers.test.ts` — all six sort tests pass a registry-style resolver; new test for the case the old map **cannot express**: a DESTRUCTIVE tool absent from any hardcoded list still sorts after every write, purely from its declared classification. New `classificationToPriority` band-ordering test.
  - `test/agent/agent-loop.test.ts` — the `buildPlugin` `getTool` stub now returns each tool's real classification (without it, every name resolved to `undefined` and the sort degraded to a no-op — caught by the existing `tool sorting` tests, which pass unchanged otherwise).

## Behavioral note

Intra-band ordering changes from the hand map's arbitrary total order (`read_file` before `list_files`, etc.) to the model's emitted order (the sort is stable). I verified no load-bearing intra-band dependency exists: cross-tool ordering inside a band — e.g. a write's target parent folder — is handled by the tools themselves (`ensureParentFolderExists` in the write tools, `fileManager.renameFile` in move), not by sort order. Band-level behavior, the actual invariant, is unchanged and now derived from the source of truth.

## Screenshots / Screencast

N/A — internal refactor; no user-visible surface (the only observable change is a new developer-facing warn log for unregistered tool names).

### Required

### AI-Generated Code

Notes on the unticked items:

- **Desktop testing**: not performed — the sort is fully covered by the unit suites (46 helper tests + 45 loop tests, including the reads-before-writes invariant tests that caught the stub-classification gap during development); the agent eval harness (`npm run eval`) exercises the full tool loop against a live vault if you want a deeper pass before merging.
- **Mobile**: no platform-specific API touched.
- **Documentation**: no docs reference `TOOL_PRIORITY` or the sort (verified by grep over `docs/`); the invariant is documented in the code comments, which moved with the code.

## Summary by CodeRabbit

* **Improvements**
  * Tool actions are now ordered according to their safety classification.
  * Read operations are prioritized before external, write, and destructive operations.
  * Unrecognized tools receive a safe fallback priority and generate a warning.
  * Ordering remains stable and supports newly introduced tool classifications.

### DIFF
--- a/src/agent/agent-loop-helpers.ts
+++ b/src/agent/agent-loop-helpers.ts
@@ -1,5 +1,6 @@
 import type { Content, Part } from '@google/genai';
 import type { ToolCall } from '../api/interfaces/model-api';
+import { ToolClassification } from '../types/tool-policy';
 import type { ToolResult } from '../tools/types';
 
 /**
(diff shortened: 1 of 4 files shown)


allenhutchison:
@coderabbitai review

PR #1476, opened 2026-09-05
fix: release headless and deleted sessions

### LINKED ISSUE #1460: lifecycle: SessionManager.activeSessions has no release path, and every hook fire leaks one entry for the plugin's lifetime
## What

`SessionManager.activeSessions` is a `Map<string, ChatSession>` that is written in three places and **deleted in none**. There is no `releaseSession` / `closeSession` / `removeSession` on the class at all. Every session the plugin has ever created or hydrated stays resident until Obsidian unloads the plugin.

For interactive sessions that is untidy but bounded by how many chats a person opens. The part that makes it a real leak is the headless path: `runHeadlessAgentTurn` calls `createAgentSession` on **every** scheduled-task fire and **every** hook fire, for a session object whose id it never exposes and whose only purpose is to carry a tool policy through one turn. Hooks are registered on `vault.on('modify')` among others, so on a vault where a `file-modified` hook matches, the map gains a permanent entry roughly per save.

Two knock-on effects, both mild individually:

- `getNoteChatSession` does `Array.from(this.activeSessions.values()).find(...)` — a linear scan that gets slower as the map fills with headless sessions that can never match it.
- `ToolExecutionEngine`'s loop-detector history is keyed by session id and has the same shape of problem. #1387 added `clearLoopDetectorSession` and wired it to exactly one caller, `SessionListModal.deleteSession`, which a headless run never reaches. So a hook that calls tools leaves detector records behind too, keyed by an id nothing will ever present again.

Routing this as an issue rather than a PR because the eviction rule is a judgment call, not a mechanism: which sessions are safe to drop, and when, differs per creation site (see below).

## Evidence

**No delete path.** `src/agent/session-manager.ts:50` declares the map; `:86` (`createNoteChatSession`), `:121` (`createAgentSession`) and `:382` (`loadSessionFromFile`) insert into it. `grep -n 'activeSessions' src/agent/session-manager.ts` returns eleven lines — three `set`, seven `get`/`values`, zero `delete`. `grep -rn 'closeSession\|removeSession\|releaseSession' src` returns nothing.

**The headless creation site** — `src/services/headless-agent-turn.ts:77-80`:

```ts
const session = await plugin.sessionManager.createAgentSession(spec.sessionLabel, {
    toolPolicy: spec.toolPolicy,
    requireConfirmation: [] as DestructiveAction[],
});
```

The function returns at `:130`, `:154` or `:166` (or throws at `:160`) and never touches the session again. Both callers — `HookRunner.runAgentTask` (`src/services/hook-runner.ts:48`) and `ScheduledTaskRunner` (`src/services/scheduled-task-runner.ts:5`) — go through it.

**Hook firing frequency** — `src/services/hook-manager.ts:636-639` registers `create`, `modify`, `delete` and `rename` vault handlers. `modify` is the one that matters: it is not a user gesture.

**The partial precedent.** `src/tools/execution-engine.ts:208-217` documents its own scope: *"Called from the session-deletion path (`SessionListModal.deleteSession`)."* That path (`src/ui/agent-view/session-list-modal.ts:342-350`) trashes the history file and calls `clearLoopDetectorSession` — and notably does **not** remove the session from `activeSessions` either, so a session the user explicitly deleted stays in the map.

**Not caught by anything.** No test asserts map size, and neither knip nor madge can see unbounded growth. `LifecycleService.onUnload` never touches `sessionManager` (`src/services/lifecycle-service.ts:204-267`), which is correct — the whole map dies with the plugin — but means the only bound is "restart Obsidian."

## Proposed unit of work

- Add `SessionManager.releaseSession(sessionId: string): void` that drops the map entry and delegates to `plugin.toolExecutionEngine?.clearLoopDetectorSession(sessionId)`, so one call covers both stores and future per-session state has an obvious home.
- Call it from `runHeadlessAgentTurn` in a `finally` around the whole turn, so it fires on the success, cancelled, exhausted and thrown paths alike. This is the case with no judgment attached — the session id never leaves the function, so nothing can look it up afterwards.
- Decide the interactive rule separately, and state it in a comment wherever it lands. The candidates: `SessionListModal.deleteSession` (the user deleted it — clearly safe, and it already clears the sibling store), and `AgentView` swapping its `currentSession` (less safe: `getRecentAgentSessions` re-hydrates from file, so an eviction here is a cache miss rather than data loss, but the view holds the object by reference and `getNoteChatSession` relies on the map for note-chat identity).
- Explicitly do **not** evict on `loadSessionFromFile`: that insert is the cache that makes re-opening a session cheap, and dropping it defeats the point.
- Add a test that a headless turn leaves the map at its pre-turn size — including the thrown-exhaustion path, which is where a naive non-`finally` release would leak.
- While in `session-list-modal.ts`, fold the existing `clearLoopDetectorSession(session.id)` call into the new `releaseSession` so there is one release verb rather than two half-releases. Keep its current placement after the successful trash (#1387's ordering rationale: if trashing failed, the session survives and its state must too).

## Out of scope

- Any TTL, LRU cap, or eviction policy for interactive sessions. The fix here is an explicit release verb called at known lifecycle points; a cache-eviction policy is a bigger design and this issue does not propose one.
- `ToolExecutionEngine`'s own execution-record retention beyond the loop detector — #1387 covered that and closed.
- The `activeSessions` linear scan in `getNoteChatSession`. It stops being a concern once the map is bounded; if it still matters afterwards, that is a separate indexing change.
- Persisting headless sessions to `Agent-Sessions/`. They deliberately never write a history file, and this issue does not change that.
- `LifecycleService.onUnload` ordering — it is correct as-is and needs no new teardown step, since the map is per-plugin-instance.

---

allenhutchison:
## Summary

Every scheduled task or agent hook retained its temporary session until plugin unload. Release that session and its tool-loop tracking after the turn completes, including cancellation and failure. Explicitly deleted chat sessions are also removed from the in-memory cache after successful trashing.

Fixes #1460

## Changes

- Add `SessionManager.releaseSession` and call it from a `finally` around the complete headless turn.
- Route successful session deletion through the same release method; preserve state if trashing fails.
- Generate new session IDs with Web Crypto UUIDs after CodeQL identified the previous `Math.random()` suffix as insecure once IDs became release keys; saved historical IDs remain valid.
- Keep interactive session caching and persisted history unchanged.
- Document temporary-session cleanup in the scheduled-task and lifecycle-hook guides.

Validation: all 4,047 tests in 179 files pass, plus format, lint, build, test typecheck, docs build, cycle detection, and Knip. Lint has 39 existing warnings and no errors. The new tests failed before the fix and cover ten headless exit paths using the real session manager, preserving an existing interactive session while releasing the temporary session. Deletion cancellation, trash failure, and session-ID uniqueness under repeated timestamps and `Math.random()` values remain covered.

Live Obsidian/eval checks were deferred at the maintainer's direction on 2026-09-06. All GitHub checks are green, and CodeRabbit approved the final head with no actionable comments.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — batch approved in the maintainer's interactive session.
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`) — full local and GitHub checks pass, including the corrected CodeQL check.
- [ ] I have tested this change on Desktop — deferred at the maintainer's direction; covered by unit and GitHub CI checks.
- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards) — no platform-specific APIs added; live mobile checks were deferred.

### AI-Generated Code

## Summary by CodeRabbit

* **New Features**
  * Temporary sessions created for automated tasks are now released after completion, cancellation, failure, or exhaustion.
  * Session identifiers now use more reliable unique UUIDs for new sessions.
  * Deleting a session also clears its associated temporary tracking state.

* **Documentation**
  * Clarified session lifecycle and cleanup behavior for scheduled tasks, lifecycle hooks, and agent mode.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -62,7 +62,7 @@ The plugin uses a factory pattern (`ModelClientFactory` in `src/api/factory.ts`)
    - **Reasoning persistence**: model "thinking" is stored on `GeminiConversationEntry.thoughts` and serialized as a collapsed `> [!reasoning]-` callout. Turns with a message (user prompt, final answer) carry a `## ` header + Message Info table; streamlined reasoning-only turns are just the bare `[!reasoning]` callout with no header. **Parser invariant** (`SessionHistory.parseHistoryContent`): entries are identified by their _callout_ (`[!user]`/`[!assistant]`/`[!reasoning]`), not a `## ` header, and the parser walks **every** callout in a `---`-delimited section (reasoning + tool callouts flow together with no dividers). This keeps the divider-free activity stream, headerless reasoning, and legacy per-entry files all round-tripping — preserve it, and cover format changes with old-format fixtures in `test/agent/session-history.test.ts`.
 5. **Custom Prompts** (`src/prompts/`): User-defined prompt templates stored in `[state-folder]/Prompts/`
 6. **Agent mode** (`src/agent/`, `src/tools/`): AI agent with tool calling capabilities
-   - Session management with persistent history
+   - Session management with persistent history; new session IDs use `session_<UUID>` from Web Crypto, while existing `session_id` values in saved history are preserved.
    - Tool registry and execution engine
    - Vault operations tools with permission system
    - Google Search integration (separate from function calling)
--- a/docs/guide/lifecycle-hooks.md
+++ b/docs/guide/lifecycle-hooks.md
@@ -6,6 +6,8 @@ Lifecycle Hooks let you trigger an AI agent run in response to Obsidian vault ev
 Hooks are disabled by default. Set **Enable lifecycle hooks** in plugin settings before any hook will fire. The default is off because vault events fire continuously and an unintentionally-broad hook can drain API quota quickly.
 :::
 
+Agent-task hooks release their temporary session and tool-loop tracking when the turn ends, including cancellation or failure. Interactive chat sessions remain available in the session browser.
+
 ## Overview
 
 A hook is a markdown file stored in `[state-folder]/Hooks/`. The file's frontmatter controls the trigger, filter, and action; the body is the prompt template.
--- a/docs/guide/scheduled-tasks.md
+++ b/docs/guide/scheduled-tasks.md
@@ -1,6 +1,6 @@
 # Scheduled tasks
 
-Scheduled tasks let you automate recurring AI prompts — daily summaries, weekly reports, periodic vault maintenance — without any manual intervention. Each task runs as a headless agent session and writes its output to a file in your vault.
+Scheduled tasks let you automate recurring AI prompts — daily summaries, weekly reports, periodic vault maintenance — without any manual intervention. Each task runs as a headless agent session and writes its output to a file in your vault. The temporary session and its tool-loop tracking are released after each run, including cancellation or failure; recurring runs do not retain one session per execution.
 
 ## Overview
 
--- a/src/agent/session-manager.ts
+++ b/src/agent/session-manager.ts
@@ -497,7 +497,7 @@ export class SessionManager {
 	 * Generate unique session ID
 	 */
 	private generateSessionId(): string {
-		return `session_${Date.now()}_${Math.random().toString(36).substring(2, 11)}`;
+		return `session_${crypto.randomUUID()}`;
 	}
 
 	/**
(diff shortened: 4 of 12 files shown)


allenhutchison:
@coderabbitai review

PR #1477, opened 2026-09-05
fix: remove ineffective system prompt override toggle

### LINKED ISSUE #1468: dead surface: the `allowSystemPromptOverride` toggle has no consumer — the safety gate it advertises does not exist
## What

`allowSystemPromptOverride` is a persisted setting with a visible toggle in **Agent Config → Custom Prompts**, a default (`false`), an i18n name and description, and a docs entry — and no reader anywhere outside the toggle that writes it. The only code that decides whether a custom prompt replaces the system prompt is `GeminiPrompts.getSystemPrompt`, which branches on the prompt's own frontmatter flag and never consults the setting.

The consequence is worse than an ordinary dead field. The toggle's description reads:

> WARNING: Allows custom prompts to completely replace the system prompt. This may break expected functionality.

A user who reads that and leaves the toggle **off** — the default — reasonably concludes that a prompt carrying `override_system_prompt: true` will be refused. It is not: the override applies regardless. The setting advertises a safety gate the code does not implement, on a switch whose whole purpose is to be a safety gate.

`docs/reference/settings.md` already documents this honestly ("**Currently has no effect**"), so this is known — but the toggle is still shipped, still enabled by default in the settings UI, and still reads as a control. Documenting a non-functional switch is not the same as resolving it, and no issue tracks the resolution.

This is also an instance of the rule in `.claude/guidelines/coding.md` — "Wiring interfaces carry only what is read" — at the settings layer rather than a context interface: the field is populated (by the toggle) but never read.

## Evidence

- `src/types/settings.ts:94` — `allowSystemPromptOverride: boolean;` declared on `ObsidianGeminiSettings`.
- `src/main.ts:94` — `allowSystemPromptOverride: false` in the defaults.
- `src/ui/settings-agent-config.ts:39-47` — the toggle: reads the value into the control, writes it back on change, saves. The only two references in `src/` outside the type declaration.
- `src/prompts/gemini-prompts.ts:138-141` — the actual decision, with no reference to the setting:
  ```ts
  if (customPrompt?.overrideSystemPrompt) {
      this.plugin?.logger.warn('System prompt override enabled. Base functionality may be affected.');
      return customPrompt.content;
  }
  ```
- `src/prompts/prompt-manager.ts:61` — `overrideSystemPrompt` comes straight from the file's `override_system_prompt` frontmatter, with no settings gate.
- `src/i18n/en.ts:633-641` — the toggle's name/description strings, including the warning quoted above.
- `docs/reference/settings.md:210-219` — documents the setting as "(legacy, currently non-functional)".

## Proposed unit of work

The decision is which way to resolve it, and that is the maintainer's call — the two directions have opposite effects on existing users:

- **Wire the gate.** Add `&& plugin.settings.allowSystemPromptOverride` (or an equivalent guard) to the branch in `gemini-prompts.ts:138`, and pass the setting through to `getSystemPrompt`'s callers. **This changes behaviour for existing installs**: every user currently relying on an `override_system_prompt: true` prompt silently loses the override on upgrade, because the default is `false`. That needs at minimum a release note, and probably a migration that flips the setting to `true` for anyone who has an overriding prompt in `[state-folder]/Prompts/`.
- **Remove the setting.** Delete the field, the default, the toggle, and the two i18n keys; rewrite the `docs/reference/settings.md` section to describe the frontmatter flag as the sole control, keeping the warning about what an override costs. No behaviour change, and the settings UI stops making a promise it doesn't keep. `src/utils/settings-migrations.ts` does not touch this key, so no migration is needed — a stale key in `data.json` is inert.

Whichever direction: cover the resulting behaviour in `test/` (there is currently no test asserting either that the setting gates the override or that it doesn't), and update `docs/reference/settings.md` plus the custom-prompts guide in the same change per the repo's documentation rule.

A reasonable default, absent other input, is **removal**: the frontmatter flag is per-prompt and explicit, a global toggle duplicates that control at a coarser grain, and removal breaks nobody.

## Out of scope

- The `override_system_prompt` frontmatter flag itself and how `PromptManager` parses it — that mechanism works and is not in question here.
- Any broader redesign of the custom-prompt system or its precedence rules against project instructions and `AGENTS.md`.
- The other settings-layer dead-field candidates: `useInteractionsApiMigrated` (read by `settings-migrations.ts`, working as intended) and `expandedSettingsSections` (UI state, read by `settings-helpers.ts`). Both were checked and are live.

---

allenhutchison:
## Summary

Remove the ineffective “Allow system prompt override” toggle and its now-empty Custom Prompts heading. The prompt's `override_system_prompt` frontmatter remains the sole control, preserving existing behavior.

Fixes #1468

## Changes

- Remove the unused settings field/default and settings UI control.
- Prune the three obsolete i18n keys across all shipped languages and translation-state entries using the existing translation serializer.
- Align the Agent Config and advanced-settings descriptions across all 20 shipped translations, and update their source hashes so the localization workflow preserves the corrected copy.
- Update README, settings reference, and the custom-prompts guide.
- Add regression coverage for the absent control/heading and for both full override and additive prompt behavior.

Validation: format, lint, build, test typecheck, all 4,037 tests in 179 files, docs build, cycle detection, and Knip pass locally and all GitHub CI checks are green. Lint has 39 existing warnings and no errors. The settings regression failed before removal; all 66 targeted settings/prompt tests pass. The translation check found 640 pre-existing stale entries before these copy fixes; unrelated translations were preserved.

## Screenshots / Screencast

Not captured. Live Obsidian verification and screenshot capture were deferred at the maintainer's direction on 2026-09-06. The removed setting row and orphaned heading are covered by the settings-render regression test. CodeRabbit reviewed the final head and reported no actionable comments.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer — batch approved in the maintainer's interactive session.
- [x] All CI checks pass (`npm test`, `npm run build`, `npm run format-check`) — full local and GitHub checks pass.
- [ ] I have tested this change on Desktop — deferred at the maintainer's direction; covered by settings-render tests and GitHub CI.
- [ ] I have verified this change does not break Mobile (or includes appropriate platform guards) — no platform-specific APIs added; live mobile checks were deferred.

### AI-Generated Code

## Summary by CodeRabbit

* **New Behavior**
  * Custom prompts replace built-in system instructions only when `override_system_prompt: true` is set in the prompt frontmatter.
  * Without this field, or when set to `false`, custom instructions are added to the built-in prompt.

* **Settings**
  * Removed the ineffective system prompt override toggle from Agent Config.

* **Documentation**
  * Updated guidance to explain frontmatter-based prompt behavior.

* **Localization**
  * Removed translations for the retired custom prompt settings.


allenhutchison:
@coderabbitai review

allenhutchison:
@coderabbitai review

allenhutchison:
@coderabbitai review

Issue #1474, opened 2026-09-05
dry-violation: OllamaModelsService and OpenAIModelsService fork one cached-model-list lifecycle


allenhutchison:
## What

`OllamaModelsService` and `OpenAIModelsService` are two independent implementations of the same policy: *fetch a provider's `/models`-style catalog over `requestUrl`, map it into `GeminiModel[]`, cache it against the identity of the endpoint it came from, and — on failure — return the previous cache only when that identity still matches, so a dead server doesn't poison the dropdown.*

That policy is subtle and load-bearing (it is what stops the model picker offering models that don't exist on the endpoint the user just switched to, and what stops an empty list sticking until a manual "Refresh"). It is currently written twice, and the copies have **already diverged** along the axis that matters — the cache key:

- Ollama keys on base URL alone (`lastBaseUrl`).
- OpenAI keys on base URL **and** API key (`lastBaseUrl` + `lastApiKey`), because a changed key must not keep serving the previous key's catalog.

Both are correct for their provider today, but the shared shape has no single home, so the next fix to the policy (a third provider, a TTL, a "stale-while-revalidate" pass, a change to what counts as endpoint identity) has to be found and applied twice — exactly the failure mode #1373 and #1350 describe on the provider clients. Two copies is below jscpd's 12-line bar in practice because the bodies differ in their identity check and their mapping, so no clone detector will catch this one.

## Evidence

- `src/services/ollama-models-service.ts:134–181` — `getModels(forceRefresh)`: identity check → early return on cache hit → `requestUrl` → `status !== 200` throw → response-shape throw → map → assign cache + `lastBaseUrl` → `logger.log('Loaded N models from <baseUrl>')` → `catch` → `logger.warn('Failed to fetch model list:')` → `return cacheMatchesBaseUrl ? (this.cachedModels ?? []) : []`.
- `src/services/openai-models-service.ts:99–142` — `getModels(forceRefresh)`: the same eleven steps in the same order, with `cacheMatches` in place of `cacheMatchesBaseUrl`, `/models` in place of `/api/tags`, `data.data` in place of `data.models`, and the same two log sentences under a different `[OpenAIModelsService]` prefix.
- `src/services/ollama-models-service.ts:184–190` and `src/services/openai-models-service.ts:148–152` — `invalidate()`: null the cached list, null every identity field, plus each service's own extra state (Ollama also clears `showCache`/`psCache` and bumps `cacheGeneration`).
[...]

PR #1558, opened 2026-09-18
fix(ollama): key showCache by endpoint and guard its write with a generation

### LINKED ISSUE #1545: Ollama's /api/show cache is keyed by model name only, so a stale probe can serve another daemon's capabilities
## What

`OllamaModelsService.probeModel()` caches `/api/show` responses in `showCache` keyed by **model name alone**, and writes the result after its `await` with no generation check:

```ts
private async probeModel(name: string, baseUrl: string): Promise<OllamaShowResponse | null> {
	if (this.showCache.has(name)) {
		return this.showCache.get(name)!;
	}
	try {
		const response = await requestUrl({ url: joinBaseUrl(baseUrl, '/api/show'), ... });
		if (response.status !== 200) return null;
		const result = response.json as OllamaShowResponse;
		this.showCache.set(name, result);   // <- no endpoint in the key, no generation guard
		return result;
	}
	...
}
```

Both clearing paths — `invalidate()` and the endpoint-change/force-refresh hook in `getModels` — empty `showCache`, but neither stops a probe that was already in flight from writing afterwards. So:

1. `getModels()` against daemon A probes `llama3.2:3b`; the `/api/show` request is pending.
2. The user retargets Ollama at daemon B (or clicks Refresh), which clears `showCache`.
3. A's probe resolves and writes its response under the bare key `llama3.2:3b`.
4. A later listing against B finds that key populated, skips probing, and takes **A's** capabilities for B's model.

The observable effect is a model on daemon B showing A's vision flag and A's trained context window. Same model name on two daemons is not exotic — it is the normal case when someone runs a local daemon and a remote one, which is exactly what the `remoteHost` / Ollama Cloud handling in this service exists for.

Note the contrast with the sibling cache in the same class: `psCache` **is** keyed on `` `${baseUrl}|${modelName}` `` and **is** generation-guarded (`getRuntimeContextLength` captures `cacheGeneration` before its probe and only writes when it still matches). `showCache` has neither. The comment on `cacheGeneration` even says "Bumped by invalidate() so a probe started beforehand can't re-seed the cache" — that protection was only ever wired to `psCache`.

## Why this is its own issue rather than part of #1474

Raised by CodeRabbit while reviewing #1544. Deliberately left out of that PR, for two reasons:

- #1474 explicitly scopes this out: *"Keep each service's provider-specific state where it is: Ollama's `showCache` / `psCache` / `cacheGeneration` and its `/api/show` + `/api/ps` probes are not part of the shared policy and should not migrate."* `probeModel` is untouched by that PR apart from routing its URL through the new `joinBaseUrl`.
- The obvious fix has a ripple. `cacheGeneration` is **shared** between `showCache` and `psCache`, so "advance the generation whenever the endpoint changes" would also begin discarding in-flight `/api/ps` results on every forced refresh — a behaviour change in a path with its own TTL and in-flight-request sharing, which deserves its own reasoning and its own tests rather than riding along in a refactor.

The bug is pre-existing and predates #1544; nothing in that PR makes it better or worse.

## Proposed unit of work

- Key `showCache` on the endpoint as well as the model name, matching `psCache`'s `` `${baseUrl}|${name}` `` shape.
- Capture a generation before the `/api/show` await and only write when it is still current, again matching `getRuntimeContextLength`.
- Decide deliberately whether the endpoint-change / force-refresh path should advance the generation, and what that should mean for in-flight `/api/ps` probes — either bump a shared generation and accept that `psCache` also drops in-flight results, or give the two caches separate generations. This is the actual design question; the keying above is the easy half.
- Cover it: a probe in flight across an endpoint switch must not populate the new endpoint's entry, and a same-named model on two daemons must get each daemon's own capabilities.

## Out of scope

- The shared cached-model-list policy in `src/services/remote-model-catalog.ts` (#1474) — the two races there are already fixed in #1544.
- Any change to what `/api/show` detects (the three-tier vision detection, context-window extraction) or to `/api/ps` runtime-window resolution; this is about cache identity only.

---

_Filed from review feedback on #1544 by the auto-dev pipeline, which kept that PR scoped rather than widening it._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DZZjscDG5fipJ24cSexoY2

allenhutchison:
## Summary

Fixes #1545

`OllamaModelsService.probeModel()` cached `/api/show` responses keyed by **model name alone** and wrote the result after its `await` with no generation check. A probe in flight across an endpoint switch (or an `invalidate()`) landed after the cache was cleared and re-seeded it under the bare name — so a later listing of the new daemon found the entry populated, skipped probing, and served the **previous** daemon's capabilities (vision flag, trained context window) for the same-named model. Same name on two daemons is the normal local-plus-remote case the `remoteHost` / Ollama Cloud handling in this service exists for.

The sibling `psCache` in the same class already had both protections (`` `${baseUrl}|${name}` `` keying and the `cacheGeneration` write guard); `showCache` had neither. The `cacheGeneration` doc comment even claimed the guard — it was only ever wired to `/api/ps`.

## The design decision (the issue's actual question)

The issue asked for a deliberate choice between bumping a shared generation and splitting it. **Split** into `showGeneration` / `psGeneration`:

- `showCache` is cleared by **two** paths — `invalidate()` *and* the endpoint-change/force-refresh `beforeFetch` hook — so its generation must advance on both.
- `psCache` is cleared **only** by `invalidate()`. A forced refresh of the *same* daemon leaves an in-flight `/api/ps` result perfectly valid — it is still that daemon's allocation. Bumping a shared generation in the `beforeFetch` hook would begin discarding good `/api/ps` results on every forced refresh, exactly the ripple the issue warned about.
- `invalidate()` bumps both.

Both caches now key through one shared `probeCacheKey(baseUrl, name)` helper (matching `psCache`'s existing shape), so the two key layouts cannot drift apart again. `probeModel` captures the generation before its `await` and writes only when it still matches, mirroring `getRuntimeContextLength`.

## Tests

Six tests in a new `show cache identity (#1545)` describe block:

- **Two mechanism tests fail against the old code** (verified by stashing the fix and re-running — exactly these two fail, red→green confirmed):
  - *suppresses a stale probe's cache write across `invalidate()`* — invalidates while the probe is past its generation capture, asserts the map stays empty. Direct verification of the guard, needed because behavioral observability of the stale write through the public API is shielded by the catalog's own clear-on-load in every reachable sequencing — which is exactly why this bug survived.
  - *keys showCache by endpoint* — same race across an endpoint switch; asserts the cache ends up holding only the new daemon's keyed entry (`http://10.0.0.1:11434|llama3.2:3b`), never the bare name or A's endpoint.
- **Four behavioral tests** covering the issue's acceptance criteria: same-named models on two daemons get each daemon's own capabilities; a probe in flight across an endpoint switch does not populate the new endpoint's entry; a probe started before `invalidate()` does not re-seed the cache; and `/api/ps` results survive a forced refresh that clears `/api/show` probes (pinning the split-generation decision).

Gates use the file's existing `let release` / `new Promise` pattern (the project's `tsconfig` lib predates `Promise.withResolvers`) and `vi.waitFor` on the recorded request rather than wall-clock sleeps.

## Out of scope (per the issue)

- The shared catalog policy in `remote-model-catalog.ts` (#1474) — untouched.
- What `/api/show` detects (three-tier vision detection, context-window extraction) and `/api/ps` runtime-window resolution — cache identity only.

## Verification

- `npm test` — 193 files / 4217 passed (4211 baseline + 6 new)
- `npm run typecheck:test` — clean
- `npm run build` — 0 errors
- `npm run lint` — 0 errors (25 pre-existing warnings elsewhere); `npm run format-check` — clean
- `npm run knip`, `npm run lint:cycles` — clean

## Screenshots / Screencast

N/A — no UI changes.

### Required

- [x] This PR is linked to an approved issue where the approach was discussed with a maintainer (#1545, filed from CodeRabbit review feedback on #1544 with the unit of work and the design question spelled out)
- [ ] I have tested this change on Desktop — N/A: cache-identity fix exercised by six new unit tests including red→green verification; no UI surface changed
- [x] I have verified this change does not break Mobile (or includes appropriate platform guards) — `requestUrl` usage unchanged; no platform-specific code touched
- [x] Documentation has been updated (if applicable) — N/A: internal cache mechanics; no user-facing surface

### AI-Generated Code

- [x] AI tool(s) used: Claude Code (oh-my-pi harness)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ollama model capability and allocation caching across daemon endpoints.
  * Prevented outdated responses from repopulating caches after endpoint changes or forced refreshes.
  * Preserved allocation results during model-detail refreshes, avoiding unnecessary repeated lookups.
  * Improved refresh reliability when multiple model checks are running simultaneously.

* **Tests**
  * Added regression coverage for endpoint-specific cache isolation, invalidation, stale responses, concurrent probes, and refresh behavior.


allenhutchison:
Both findings verified against the code and fixed in 91c758c6:

- **Outside-diff (psInFlight, valid):** the two halves go together — storing the generation per entry and reusing it only when current means a stale probe and a fresh probe for the same key can now coexist, so the finally must remove the entry only if it still holds this probe (otherwise the stale probe evicts the newer one). Added a regression test for exactly that (red against the old bare-delete semantics).
- **Nitpick (test, valid):** rewrote the ps-isolation test so the /api/ps probe stays pending across getModels(true). If the split-generation design ever regressed to a shared generation, the pending write is suppressed and the final lookup re-probes — verified red by temporarily bumping psGeneration in the beforeFetch hook.

Gates: 193 files / 4218 tests, typecheck:test, build, lint, format-check, knip, lint:cycles — all green.

Issue #1499, opened 2026-09-11
invariant drift: the tool sort reorders execution but not the replayed functionCall turn, so calls and results can pair up wrong


allenhutchison:
## What

`AgentLoop` sorts each tool batch so reads run before writes/deletes (the tool-execution-ordering
invariant, #1424), executes the **sorted** array, and then builds the follow-up history from the
**unsorted** array. The model turn's `functionCall` parts therefore come out in the model's emitted
order while the matching `functionResponse` parts come out in execution order. Whenever a batch
contains mixed classifications emitted out of priority order — which is the only case the sort exists
for — the two turns disagree.

When the model supplies correlation ids (Interactions `call_id`, OpenAI `tool_call_id`, #1398) the
pairing is by id and the order is cosmetic. On plain `generateContent` there are no ids, and on the
Interactions path a model that omits them falls back to the **tool name** as the id — so two parallel
calls of the same tool collapse onto one `call_id` and there is nothing left to disambiguate them but
position, which is exactly what the sort has just changed. The model then reads a `read_file` result
as the answer to its `delete_file` call, or gets two same-named results in the wrong order.

## Evidence

- `src/agent/agent-loop.ts:349` — `const sortedToolCalls = sortToolCallsByPriority(currentToolCalls, …)`
  returns a **new** array; `currentToolCalls` keeps the model's emitted order.
- `src/agent/agent-loop.ts:359` — `executeToolBatch(sortedToolCalls, …)`; the batch loop pushes one
  result per call in `sortedToolCalls` order (`src/agent/agent-loop.ts:683–748`), so `toolResults` is
  in execution order.
- `src/agent/agent-loop.ts:416–422` — the live follow-up path builds history with
  `toolCalls: currentToolCalls, toolResults` — the two differently-ordered arrays. This history is
  what goes back to the model at `src/agent/agent-loop.ts:466+`.
- `src/agent/agent-loop.ts:373–379` — same mismatch on the loop-abort path.
- `src/agent/agent-loop-helpers.ts:184–185` — `buildToolHistoryTurns` emits the model turn from
  `buildFunctionCallParts(toolCalls)` and the user turn from `buildFunctionResponseParts(toolResults)`
  with no reconciliation between them.
- `src/api/providers/gemini/interactions-mapper.ts:83` and `:92` — `id: part.functionCall.id ??
  part.functionCall.name ?? 'call'` / `call_id: part.functionResponse.id ?? part.functionResponse.name
[...]

PR #1593, opened 2026-09-21
refactor(agent): decompose the two remaining Tier 1 oversized methods (#1396)

### LINKED ISSUE #1396: weak abstraction: 26 methods run past 120 lines, with three outliers over 350 — a cohesion axis #801 doesn't measure
## What

`src/` has 26 methods of 120 lines or more, totalling ~5,360 lines — roughly 8% of the non-i18n source sitting inside two dozen function bodies. Three are outliers by a wide margin: `createInputArea()` at 533 lines, `sendMessage()` at 505, and `AgentLoop.run()` at 351.

This is a **different axis from #801**, which tracks oversized *files* and is currently parked (`auto:skip`). A file can be a reasonable size and still hold a 500-line method, and splitting a file does not by itself shorten the methods inside it — `agent-view-send.ts` is 859 lines of which 505 are one method, so "split the file" has no obvious seam to cut along until that method is decomposed first. Two of the three outliers (`agent-view-send.ts`, `agent-loop.ts`) are not on #801's file list at all. Filing this separately so the method-length question can be answered on its own terms rather than inheriting #801's parked status.

Why it matters here specifically: `sendMessage()` and `run()` are where this repo's load-bearing invariants live — the frozen turn timestamp and byte-stable `perTurn` fields for prefix-cache alignment, the `protectFromIndex` pin that keeps mid-loop compaction from folding in-flight `functionCall`/`thoughtSignature` continuity into a summary, the deliberate lazy `require('./agent-factory')`. Those constraints are documented in `.claude/guidelines/invariants.md` and in comments inside the method bodies, but a 500-line body is where an invariant goes to be quietly violated by an unrelated edit, because no reviewer holds the whole thing in their head during a diff.

Measured on `ae08f7d`, tab-indented method declarations at class depth, counting to the closing `\t}`. Loop/conditional keywords excluded.

## Evidence

**Tier 1 — outliers (>350 lines), worth their own sub-issue each**

- `src/ui/agent-view/agent-view-ui.ts:344` — `createInputArea()`, 533 lines. Pure DOM construction; the longest single body in the repo.
- `src/ui/agent-view/agent-view-send.ts:232` — `sendMessage()`, 505 lines. Nine identifiable phases run end to end in one body: guard + input harvest (232–262), attachment persistence and the send→stop button transition (264–290), display-message assembly from saved attachments (292–336), request-context assembly — history snapshot, context files, custom prompt, project instructions, `additionalInstructions`, tool discovery (338–436), compaction + request build (438–493), plan-mode approval (498–521), the streaming arm (523–661), the non-streaming arm (662–704), and error/`finally` (705–735). PR #1395 removes ~16 lines of duplication between the two arms but deliberately does not touch the shape.
- `src/agent/agent-loop.ts:247` — `run()`, 351 lines. Highest-risk of the three: it carries the `protectFromIndex` compaction pin, the loop-abort threshold, the empty-response retry, cancellation polling, and the lazy `require`. Any split has to preserve all of them.

**Tier 2 — 200–265 lines**

- `src/ui/agent-view/agent-view-tool-display.ts:412` — `showToolResult()`, 265
- `src/services/rag-vault-scanner.ts:386` — `_doIndexVault()`, 240
- `src/ui/mcp-server-modal.ts:67` — `onOpen()`, 235
- `src/ui/agent-view/agent-view-ui.ts:113` — `createCompactHeader()`, 227
- `src/ui/hook-management-modal.ts:198` — `renderFormBody()`, 224
- `src/ui/agent-view/session-settings-modal.ts:27` — `onOpen()`, 223
- `src/ui/agent-view/agent-view-messages.ts:746` — `displayConfirmationRequest()`, 219
- `src/ui/agent-view/agent-view-tools.ts:93` — `handleToolCalls()`, 202

**Tier 3 — 120–200 lines (14 methods)**

`agent-view-messages.ts:495` `showEmptyState()` (192) · `agent-view.ts:213` `createAgentInterface()` (154) · `web-fetch-tool.ts:56` `execute()` (154) · `scheduler-management-modal.ts:210` `renderFormBody()` (152) · `vault/read-file-tool.ts:47` `execute()` (149) · `agent-view-messages.ts:86` `displayMessage()` (147) · `vault/search-file-contents-tool.ts:51` `execute()` (146) · `grounding-tool-runner.ts:34` `getChunkCitation()` (137) · `context-manager.ts:425` `prepareHistory()` (137) · `providers/gemini/client.ts:340` `generateStreamingResponse()` (136) · `execution-engine.ts:41` `executeTool()` (135) · `vault-analyzer.ts:47` `initializeAgentsMemory()` (130) · `rag-search-tool.ts:139` `execute()` (126) · `scheduler-management-modal.ts:84` `renderEntityRow()` (123) · `providers/ollama/client.ts:94` `generateStreamingResponse()` (120)

## Proposed unit of work

- **Settle the threshold first, before any code moves.** Decide whether 120 is the right bar for this codebase and whether Tier 3 is debt at all — several Tier 3 entries (the `execute()` methods, `renderEntityRow()`) are linear validate-then-act bodies where splitting would add indirection without adding clarity. A plausible outcome is "Tier 1 and 2 only, leave Tier 3 alone," and recording that decision is worth as much as the refactor.
- **One sub-issue per Tier 1 method**, each a pure code-motion refactor. Suggested seams:
  - `sendMessage()` — extract the request-context assembly (338–436) into a `buildTurnRequest(...)` returning the `ExtendedModelRequest` plus `perTurn`, and lift the two API arms into `runStreamingTurn(...)` / `runNonStreamingTurn(...)`. The `perTurn` object already exists precisely because those fields must stay byte-stable across the initial call and every follow-up — it is the natural parameter object for the extraction.
  - `createInputArea()` — split by widget group (input surface, attachment shelf controls, send/stop + plan-mode buttons), following the `agent-view-*.ts` module convention already in use.
  - `run()` — extract the per-iteration body (tool batch → history construction → follow-up request) from the iteration/termination scaffolding, keeping every invariant listed above intact and the lazy `require` where it is.
- **Tier 2 is opportunistic**: take one when already editing the file, not as a campaign.
- Cover each split with the file's existing test file rather than new tests — these are extractions. `agent-loop.ts` and `context-manager.ts` have the strongest existing nets; `agent-view-ui.ts` and `mcp-server-modal.ts` have none, which is #1262's problem, not this one's.
- If a bar is agreed, consider a mechanical guard so the tier can't quietly regrow — same argument as #1264 makes for duplicate blocks. That would be its own issue.

## Out of scope

- **File-level splits.** That is #801, and it is parked. This issue is only about method length; a split that shortens methods may well leave the files the same size, and that is fine.
- **Behavior changes of any kind.** Every extraction here must be observably identical, including the streaming/non-streaming ordering differences and the invariants named above. If a split surfaces what looks like a latent bug, file it separately rather than correcting it inside a code-motion PR.
- **PR #1395**, which removes the duplicated usage-metadata emit between `sendMessage()`'s two arms. It is a dedup, not a decomposition, and lands independently.
- **The i18n files** (`src/i18n/*.ts`), excluded from the measurement — they are generated data tables.
- Adding test coverage for the untested UI modules involved (#1262).
- Choosing or configuring a CI guard for method length — a follow-up if and only if a threshold is agreed.

---

allenhutchison:
## Motivation

Closes out the Tier 1 work on #1396: the two remaining oversized methods are decomposed into private phase methods on the same class, **pure code motion, no behavior change** — same approach as the merged `createInputArea()` slice (#1465).

- `sendMessage()` (`src/ui/agent-view/agent-view-send.ts`): **479 → 13-line orchestrator**
- `AgentLoop.run()` (`src/agent/agent-loop.ts`): **390 → 256-line state machine**

### `sendMessage()` — six extractions
- `captureTurnInput()` — frozen turn timestamp + preamble. **The timestamp/preamble move as one piece with their consumers**: the timestamp is returned and threaded into `assembleTurnRequest` → `addEntryToSession` so the persisted `| Time |` row still matches the preamble the model saw (cache alignment); nothing recomputes or re-derives it.
- `beginExecutionUi()` — executing flag + send→Stop button swap.
- `buildDisplayMessage()` — attachment-preview display copy (image embeds, source callouts).
- `assembleTurnRequest()` — history snapshot, user-entry persistence, context files, custom prompt, project instructions, tool discovery, pre-turn compaction, request + `perTurn` build, plan-mode approval. Returns a `TurnRequest` (module-level interface) so `perTurn` is built once and threaded, not re-derived.
- `buildAdditionalInstructions()` — the per-turn context notes.
- `dispatchResponse()` — streaming/non-streaming arms; routes to `handleToolCalls` or `finalizeNoToolCallResponse` exactly as before.

### `AgentLoop.run()` — five extractions
- `sortBatchForExecution()` — classification-derived priority sort (#1424) with the sourceIndex pre-tag (#1499).
- `checkBatchEscalations()` — loop-detector `AGENT_LOOP_ABORT_THRESHOLD` abort + `stopOnToolError` end-of-turn.
- `buildPostBatchHistory()` — tool-response turn build + mid-loop compaction. **The `protectFromIndex` pin, the post-compaction `loopStartIndex` shift, and the double cancellation check (before and after `prepareHistory`) all carried over verbatim** and return via a discriminated union.
- `requestFollowUp()` — streaming follow-up (chunk hook + `onFollowUpStreamReady` register/clear pair) and the intermediate `onModelReasoning` hook, in the original order.
- `retryAfterEmpty()` — the empty-response retry request + fallback return.

**Explicitly preserved:** the lazy `require('./agent-factory')` stays inside `run()` (import-cycle invariant, invariants.md line 67); the `perTurn` byte-stable threading; the loop-abort threshold accounting; the degenerate no-tool-calls return. Every invariant comment is moved verbatim, not rewritten.

## Verification

- `test/ui/agent-view/agent-view-send.test.ts` (14 tests) and `test/agent/agent-loop.test.ts` (46 tests) pass **unchanged — zero assertions edited**, the gate for a code-motion PR.
- Full pre-flight green: `format-check`, `lint` (0 errors; same 27 pre-existing warnings as master), `build`, `typecheck:test`, `test` (4267 passed), `lint:cycles` (0 cycles), `knip` (clean).

⚠️ **Behavioral verification in a live vault did not run.** As with #1465, the sandbox has no Obsidian runtime. The outstanding manual pass should now cover both slices: vault `.md` drop, vault `.png` drop, external image drop, paste, `@`, `/`, Enter-to-send, Stop mid-stream, send with attachments only, a turn triggering pre-turn compaction, and plan-mode approve/reject.

## Docs

No docs update — no user-visible behavior, setting, or public-surface change; all extracted methods are private. `AGENTS.md` describes the send path and loop invariants at a level this doesn't change, and the invariant comments moved verbatim with their code.

Closes #1396 (Tier 1 complete; see the closing comment there for the threshold decision).

## Summary by CodeRabbit

- **Improvements**
  - Improved agent responses when tool calls fail, including clearer failure handling and fallback messages.
  - Enhanced reliability for follow-up requests, empty responses, cancellation, and long conversations.
  - Preserved tool execution order and improved handling of unrecognized tools.
  - Improved message sending with more consistent streaming and non-streaming behavior.
  - Enhanced attachment handling and usage information during requests.
  - Plan-mode rejections now clearly indicate that the turn was aborted.


allenhutchison:
## Live-vault verification — done (Test Vault, plugin `gemini-scribe` 4.11.0, bundle md5-verified identical to this PR's build)

(Supersedes my previous comment, which lost inline code to shell interpolation — this one is the full record.)

Covers the outstanding manual pass for **both** #1465 and this PR. Zero console errors across the whole pass (`obsidian dev:errors` clean at every checkpoint).

| Check | Result |
|---|---|
| Enter-to-send | ✅ user + model messages rendered, state reset |
| Send→Stop button transition | `beginExecutionUi` swap to Stop (`aria-label="Stop agent execution"`, `gemini-agent-stop-btn` class) and reset after completion — both observed |
| Model response + reasoning | Model replied `REFACTOR-OK`; reasoning section rendered |
| Session persistence + frozen timestamp | the persisted `| Time |` row (`2026-09-21T16:04:14.443-07:00`) **matches the `[Current date and time: …]` preamble byte-for-byte** — the cache-alignment invariant holds on the extracted `captureTurnInput`/`assembleTurnRequest` path |
| Paste image | Generated PNG pasted → `gemini-shelf-item-image` thumbnail on shelf |
| Attachment send | Attachment persisted to `attachments/pasted-image-*.png`, Image Source callout appended by `buildDisplayMessage`, model received and described the inline image |
| Vault `.md` drop (wikilink) | Text context chip created; follow-up turn answered with the file name — `buildFileContext` + the CONTEXT FILES note reached the model |
| External (filesystem) image drop | File-object drop routed to the image shelf item |
| `@` mention picker | Suggestion items rendered visibly |
| Plan mode | Toggle → active state; plan produced, approval UI shown; Approve → "Proceed with the approved plan." → executed and finished; plan-mode flag cleared after the turn (`conductPlanApproval` `finally` verified) |
| Stop mid-turn | Stop clicked during execution → "Agent execution canceled" notice, `isExecuting` back to `false` |

**Live `AgentViewSend` method table confirmed** the extraction shipped: `captureTurnInput`, `beginExecutionUi`, `buildDisplayMessage`, `assembleTurnRequest`, `buildAdditionalInstructions`, `dispatchResponse` all present on the prototype alongside the pre-existing methods.

Caveat for honesty: OS-level drag events can't be synthesized from the page context, so the drop checks dispatch `DragEvent`s with `DataTransfer` payloads (File objects for the filesystem path, a wikilink for the vault path) — the handlers' real code paths, but not a literal pointer drag. Slash-command (`/`) picker did not open on a bare `/` keydown in this harness; the keydown path is suite-covered and the mention picker on the same handler fired, but flagging it in case you want to eyeball that one by hand.

